PR #208, opened 2014-08-29
Update HtmlToolbarHelper.php

skalmi:
Try to make array from objects
This is useful for cases like
```php
class Date{
	private $time;
	public function setTime(){
		$this->time = time();
	}
}
class Name{
	private $first;
	public function setName(){
		$this->first = 'a1';
	}
}
$Art = new Object();
$Art->date = new Date();
$Art->name = new Name();
$Art->date->setTime();
$Art->name->setName();
```


dereuromark:
@skalmi Sure this is still needed? Or can we close?

PR #292, opened 2015-01-01
Alternate background color for debug tables

kamui545:
Hi guys,

I added alternative style for odd rows to improve debug tables usability.

I had to rewrite the environments and SQL log tables to avoid some CSS problems due to the DOM (no thead/tbody tags). I hope it's not a problem that I removed the use of the HTML helper to construct those tables, I think it's more readable/flexible this way and more consistant regarding the tables from the others panels.

### DIFF
--- a/src/Template/Element/cache_panel.ctp
+++ b/src/Template/Element/cache_panel.ctp
@@ -22,13 +22,16 @@
         <span class="inline-message"></span>
         <table cellspacing="0" cellpadding="0" class="debug-table">
             <thead>
-                <tr><th><?= __d('debug_kit', 'Metric') ?></th><th><?= __d('debug_kit', 'Total') ?></th></tr>
+                <tr>
+                    <th><?= __d('debug_kit', 'Metric') ?></th>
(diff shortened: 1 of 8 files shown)


lorenzo:
Could you paste a screenshot after the changes?

lorenzo:
It looks like you changed the indentation style of the files, so it is difficult to see the real diff.

markstory:
I thinl removing the helper usage is fine. It might be best to remove it in general, as I suspect there aee not many calls to that method anymore.

kamui545:
screen shot 2015-01-02 at 00 33 27

Yeah I fixed some bad indentation as well.
And a syntax error in the CSS file (using // as comment instead of /* */)

kamui545:
I removed the method to construct the table as it's not used anymore

markstory:
Looks like there were tests for the removed method.

kamui545:
Indeed, I forgot to run PHPUnit, my bad. I'll fix it

kamui545:
Everything is green :+1:

markstory:
Thanks @kamui545

PR #293, opened 2015-01-01
Fix to avoid double scrollbars

kamui545:
Hide ```<body>``` overflow when the toolbar is expanded (100% height) to avoid the double scrollbars (only keep the scrollbar from the iframe).


lorenzo:
Thanks @kamui545 !

PR #297, opened 2015-01-05
Update EnvironmentPanel.php

DIDoS:
Change FULL_BASE_URL to Router::fullbaseUrl() since it is Deprecated

http://book.cakephp.org/2.0/en/core-libraries/global-constants-and-functions.html#FULL_BASE_URL


ravage84:
This produces errors with older CakePHP versions, e.g. 2.3.10:
> Fatal Error Error: Call to undefined method Router::fullbaseUrl() in [/home/travis/build/cakephp/cakephp/plugins/DebugKit/Lib/Panel/EnvironmentPanel.php, line 52]

markstory:
Given that the constant won't go away in 2.x I'd prefer to have debug kit working across the board.

PR #303, opened 2015-01-27
Periodically clear out old data.

markstory:
To combat gigantic debugkit databases, we can periodically clear out old data. There may be situations where we end up clearing out all but the most recent request, but I think that's acceptable. If a person hasn't interacted with an app in 2+ weeks, they are unlikely to remember what they last did.


lorenzo:
Good job! :D

PR #311, opened 2015-03-02
Use dot-notation when loading plugin panels

deizel:
Otherwise `App::className()` can't resolve the path to the panel.


markstory:
Thanks :+1:

PR #316, opened 2015-04-01
Scrolling toolbar on small screen

planardothum:
In responsive design the variables tab gets lost from the toolbar when the screen is small. 
This solution is not elegant but it saves wondering "where did my tabs go" on smaller screen sizes.
Some sort of media-query driven popup or horizontal scroller would be better in the long run. Maybe that is in the works?

### DIFF
--- a/webroot/css/toolbar.css
+++ b/webroot/css/toolbar.css
@@ -62,7 +62,7 @@ p {
 }
 .toolbar.open {
 	border-radius: 0;
-	overflow: scroll;
+	overflow: auto;
 }
 
 /* Open close button */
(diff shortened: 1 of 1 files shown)


markstory:
There aren't any plans to make debugkit responsive right now. But this is a good temporary solution.

webroot/css/toolbar.css
```diff
@@ -62,6 +62,7 @@ p {
 }
 .toolbar.open {
 	border-radius: 0;
+	overflow: scroll;
```
markstory: Won't this always force scrollbars even when there is no overflow?

planardothum:
@markstory you're right. I only tested it in chrome on Mac where you never see scrollbars. Changed to auto.

markstory:
Thanks :+1:

PR #321, opened 2015-04-14
Add missing `/`, fixes tests on sqlite

jadb:


ceeram:
Makes the tests at least run for me, so good job on that one. Still getting another fatal though locally.
Here is the output of phpunit of both master and of this PR:

https://gist.github.com/ceeram/020b3c102f196c29790c

Haven't investigated this error further yet, `General error: 17 database schema has changed`

ceeram:
the DROP TABLES in RequestTableTest seem to cause the schema error. If run all tests without RequestTableTest then all passes.

tests/bootstrap.php
```diff
@@ -88,7 +88,7 @@
 
 // Ensure default test connection is defined
 if (!getenv('db_dsn')) {
-    putenv('db_dsn=sqlite://' . TMP . 'debug_kit_test.sqlite');
+    putenv('db_dsn=sqlite:///' . TMP . 'debug_kit_test.sqlite');
```
markstory: TMP normally starts with `/`

ceeram: according to php.net only a single / would do, but the gist shows it really needs the extra /

markstory: Ok.

ceeram: In fact, putenv('db_dsn=sqlite:/' . TMP . 'debug_kit_test.sqlite'); would have worked too.
So it works with 1 or 3 slashes, just not with 2

markstory: Its like the holy hand grenade from monty python.

jadb: I used 3 like in the tests: https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Core/StaticConfigTraitTest.php#L222

PR #331, opened 2015-04-25
Implemented Ajax planel, closes #261

### LINKED ISSUE #261: RFC - toolbar should reload periodically or on Ajax complete
Given that firelog is gone, the toolbar should listen for sent Ajax requests or reload itself periodically in order to show data collected from xhr requests.

lorenzo:
This is implements the Ajax panel proposed in #261 

ajax_panel


zoghal:
:100: :+1:

jadb:
:+1:

markstory:
Clever stuff.

markstory:
Looks great, would the interface be simpler/clearer if instead of going into a separate a separate panel Ajax requests appended to the top of the history panel? They could be flagged as xhr requests too if visually separating them is important.

webroot/js/toolbar-app.js
```diff
@@ -221,10 +222,21 @@ Toolbar.prototype = {
 		}
 	},
 
+	onMessage: function(event) {
+		if (event.data.startsWith('ajax-completed$$')) {
+			this.ajaxRequests.push(JSON.parse(event.data.split('$$')[1]));
+		}
+    },
```
ADmad: Indentation seems off.

ADmad:
Reusing history panel and color coding ajax request would be nice indeed.

jameswatts:
Very nice!

lorenzo:
@markstory I would personally would find it more difficult to understand finding ajax requests in history. They would also be a bit more difficult to separate visually and to prevent from overflowing the history panel.

I'm open to ideas on how to present them visually if we are to put them in the history panel...

lorenzo:
Added the `Content-Type` to the request information and now request are listed showing newest first

lorenzo:
@markstory @ADmad This is what I came up with for merging the two panels
ajax panel

webroot/js/toolbar-app.js
```diff
@@ -221,10 +222,26 @@ Toolbar.prototype = {
 		}
 	},
 
+	onMessage: function(event) {
+		if (event.data.startsWith('ajax-completed$$')) {
```
markstory: I don't think startsWith is available everywhere yet.

markstory:
@lorenzo Looks fantastic!

ADmad:
:clap:

lorenzo:
Going to merge and tag a new release, thanks for your feedback :)

PR #332, opened 2015-04-30
js_debug_toolbar.js activatePanel use show()

zeroasterisk:
Changed `js_debug_toolbar.js` `activatePanel()` to jQuery `show()`

Rationale: This makes sense because `deactivatePanel()` is using jQuery `hide()`

Justification: If you're going to use hide, you should toggle with show. -- If you aren't going to use show, you shouldn't use hide.

Discovery: This was found because on a project show/hide were over-ridden to set/remove classes vs. simply altering the display property....


markstory:
Thanks :+1:

PR #335, opened 2015-05-10
Simplify data serialization through wrapper function

josegonzalez:
Previously, we would not surface *which* variable had errors serializing, resulting in an empty `variables` panel when any data was unserializable. With this change, variables that cannot be serialized are noted as such in the data array, as well as in the `error` key.

### DIFF
--- a/src/Routing/Filter/DebugBarFilter.php
+++ b/src/Routing/Filter/DebugBarFilter.php
@@ -191,13 +191,7 @@ class DebugBarFilter extends DispatcherFilter
 
         foreach ($this->_registry->loaded() as $name) {
             $panel = $this->_registry->{$name};
-            try {
-                $content = serialize($panel->data());
-            } catch (\Exception $e) {
-                $content = serialize([
-                    'error' => $e->getMessage(),
-                ]);
-            }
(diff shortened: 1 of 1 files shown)


src/Routing/Filter/DebugBarFilter.php
```diff
@@ -213,6 +207,40 @@ public function afterDispatch(Event $event)
...
+                try {
+                    serialize($value);
+                    $response[$key] = $value;
+                } catch (\Exception $e) {
+                    $response['error'] = $e->getMessage();
+                }
+            }
+        }
+
+        return serialize($response);
```
markstory: Given that only the variables panel is the problem, couldn't this go there i stead of being applied to all panels?

josegonzalez: Probably. How do you propose I handle multiple errors?

markstory: Each problematic top level view var could have its content replaced with an inline error perhaps.

josegonzalez: Alright, I'll take a look at that.

josegonzalez:
Closed by #337 I think.

PR #336, opened 2015-05-11
remove object inspection of last commit

Graziel:
f9f615c breaks the variable inspection

these lines breaks my ajax calls and gives 
```
You cannot serialize or unserialize PDO instances
```
on some variables

### DIFF
--- a/src/Panel/VariablesPanel.php
+++ b/src/Panel/VariablesPanel.php
@@ -95,9 +95,6 @@ class VariablesPanel extends DebugPanel
                     $item->getLine()
                 );
             }
-            if (is_object($item) && method_exists($item, '__debugInfo')) {
-                $item = $item->__debugInfo();
-            }
             return $item;
         });
 


markstory:
Which object contained the pdo instance? Using debgInfo should let us avoid those errors.

Graziel:
this is my setup its nothing exceptional

LanguagesTable
```
<?php
namespace App\Model\Table;

use Cake\ORM\Table;

class LanguagesTable extends Table
{
}
```

in AppController beforeFilter i call

```
$languages = TableRegistry::get('Languages');

$this->languages = $languages->find()
            ->where(['NOT' => ['Languages.id' => 1]])
            ->all();

$this->set('all_languages', $this->languages); // <- here, if i comment it debug kit shows other variables that are ok
```

markstory:
Ok, that's good to know, I'll take a look as query's have a __debugInfo method.

PR #340, opened 2015-05-19
Make it possible for non-plugin panels to work.

markstory:
Previously the element name resolution would fail even when `$plugin = false`.


markstory:
When writing the docs for this feature I came to the conclusion that non-plugin panels didn't actually work.

antograssiot:
:+1:

PR #344, opened 2015-05-30
Update readme to clone the right version on Cake 2.x

jrbasso:
The original instructions make people to download the latest version, which is target to Cake 3.x


dereuromark:
:+1:

Issue #345, opened 2015-06-01
sort variables


Graziel:
i find that having many viewVars can be annoying to find in debugKit variables, it would be easy to implement ksort after https://github.com/cakephp/debug_kit/blob/master/src/View/Helper/ToolbarHelper.php#L77-L79 as simple
```
if (is_array($values)) {
      ksort($values);
}
```
downside is some users can rely on current setting of non ordered results
(ie to check the ->order in ->find('list') )

markstory:
We could sort just the top level variable scope and not recursively through the variables.

dereuromark:
@markstory I am hesitant to upvote that because the order of the variables usually tells you when it was added. Mixing that up might lose that implicit information...

ravage84:
:+1:  to JS sorting

markstory:
Good point @dereuromark knowing the approximate insert order is handy.

Graziel:
another approach would be to add button which adds cookie via js, reloads the page and change condition to 
```
if (is_array($values) && $cookie_sort_set) {
      ksort($values);
}
```
this way it will remember it so you wont have to click every time you load panel to sort via js and if you dont want sort just delete the cookie again via js and refresh the panel

dereuromark:
Sounds like a good idea to me.

markstory:
I don't think I understand this well enough to build it. Someone will have to do a pull request :smile:

markstory:
Closing as there is a pull request open.

PR #346, opened 2015-06-02
added sort to variables panel

Graziel:


ravage84:
Refs https://github.com/cakephp/debug_kit/issues/345

webroot/js/toolbar-app.js
```diff
@@ -130,6 +130,17 @@ Toolbar.prototype = {
 	},
 
 	bindNeatArray: function() {
+        var sort_button = this.content.find('.neat-array-sort');
+        var _this = this;
+        sort_button.click(function() {
+            if ($(this).attr('checked')) {
+                document.cookie = 'debugKit_sort=; expires=Thu, 01 Jan 1970 00:00:01 GMT; path=/';
```
markstory: Doesn't this run the risk of erasing all the other cookies?

webroot/js/toolbar-app.js
```diff
@@ -130,6 +130,17 @@ Toolbar.prototype = {
 	},
 
 	bindNeatArray: function() {
+        var sort_button = this.content.find('.neat-array-sort');
```
markstory: variables should be camelBack.

webroot/js/toolbar-app.js
Graziel: http://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-8747038

"When this attribute is set, the value it is set to should be a string that adheres to the cookie non-terminal of [IETF RFC 2965]; that is, it should be a single name-value pair followed by zero or more cookie attribute values."

webroot/js/toolbar-app.js
```diff
@@ -130,6 +130,17 @@ Toolbar.prototype = {
 	},
 
 	bindNeatArray: function() {
+        var sortButton = this.content.find('.neat-array-sort');
+        var _this = this;
+        sortButton.click(function() {
+            if ($(this).attr('checked')) {
+                document.cookie = 'debugKit_sort=; expires=Thu, 01 Jan 1970 00:00:01 GMT; path=/';
```
ADmad: Instead of '/' the path should be app's base url.

Graziel:
anything else to change?

src/Template/Requests/view.ctp
```diff
@@ -38,7 +38,8 @@ $(document).ready(function() {
         panelClose: $('#panel-close'),
         keyboardScope : $(document),
         currentRequest: '<?= $toolbar->id ?>',
-        originalRequest: '<?= $toolbar->id ?>'
+        originalRequest: '<?= $toolbar->id ?>',
+        baseUrl: '<?= $this->Url->build('/') ?>'
```
markstory: Should json_encode this value. Someone somewhere will have ' in their basedir.

src/View/Helper/ToolbarHelper.php
```diff
@@ -77,6 +88,9 @@ public function makeNeatArray($values, $openDepth = 0, $currentDepth = 0, $doubl
         if (empty($values)) {
             $values[] = '(empty)';
         }
+        if ($this->sort && is_array($values)) {
+            ksort($values);
+        }
```
markstory: This will apply to each nested object as well. Is that what is intended?

webroot/js/toolbar-app.js
```diff
@@ -130,6 +131,17 @@ Toolbar.prototype = {
...
+        var sortButton = this.content.find('.neat-array-sort');
+        var _this = this;
+        sortButton.click(function() {
+            if ($(this).attr('checked')) {
+                document.cookie = 'debugKit_sort=; expires=Thu, 01 Jan 1970 00:00:01 GMT; path=' + _this.baseUrl;
+            } else {
+                document.cookie = 'debugKit_sort=1; path=' + _this.baseUrl;
+            }
+            _this.loadPanel(_this.currentPanel());
+        });
```
markstory: The indentation is inconsistent. The rest of this file is tabs. There is probably a good case for converting the file over to spaces, but that can be done separately I guess.

src/View/Helper/ToolbarHelper.php
Graziel: yup it was, but if you think it should do it only on top level i can change it or even add another checkbox for deep sorting

markstory: I don't know if more checkboxes is a great experience. The one checkbox should be fine for now.

markstory:
Thanks @Graziel merged and tagged 3.1.5

PR #347, opened 2015-06-12
Adding the version as title to the button

dereuromark:
Adding the version as title to the panel button to quickly know the current running cakephp version, showing

    CakePHP 3.0.6 Debug Kit

on hover now.
Also useful for beginners or lazy ones who quickly want to know the current version the app is running with.

### DIFF
--- a/src/Template/Requests/view.ctp
+++ b/src/Template/Requests/view.ctp
@@ -1,5 +1,7 @@
 <?php
 use Cake\Routing\Router;
+use Cake\Core\Configure;
+
 ?>
 <div id="panel-content-container">
     <span id="panel-close" class="button-close">&times;</span>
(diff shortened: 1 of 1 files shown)


ravage84:
*Could* be misunderstood for the DebutKit version, though.

dereuromark:
Not really, then it would be `CakePHP Debug Kit 3.x.x`.

dereuromark:
Note the `Welcome to CakePHP v3.0.6 Console` message for CLI :) Which no one would interpret differently, either.

src/Template/Requests/view.ctp
```diff
@@ -22,7 +24,8 @@ use Cake\Routing\Router;
     </li>
     <?php endforeach; ?>
     <li id="panel-button">
-        <?= $this->Html->image('DebugKit.cake.icon.png', ['alt' => 'Debug Kit']) ?>
+        <?= $this->Html->image('DebugKit.cake.icon.png',
+			['alt' => 'Debug Kit', 'title' => 'CakePHP ' . Configure::version() . ' Debug Kit']) ?>
```
markstory: The CakePHP version is already show in the Environment panel, why duplicate that here?

dereuromark: It is quite hidden there, though, you have to scroll a quite a bit to find it.

markstory:
Ok. Lets give this a shot and see what people think.

PR #351, opened 2015-06-21
Better output of timer content.

dereuromark:
Currently, the

    Timer ( 0,25 s - 9,56 MB )

output reads as range by default. If you quickly look at it the `-` dash makes you think it is some range.
Using `/` is probably a better idea:

    Timer ( 0,25 s / 9,56 MB )

Same for the SQL Log (`Sql Log ( 0 - 0 ms )`)

### DIFF
--- a/src/Panel/SqlLogPanel.php
+++ b/src/Panel/SqlLogPanel.php
@@ -92,6 +92,6 @@ class SqlLogPanel extends DebugPanel
             $count += count($logger->queries());
             $time += $logger->totalTime();
         }
-        return "{$count} - $time ms";
+        return "$count / $time ms";
     }
 }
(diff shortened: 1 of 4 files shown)


dereuromark:
I also added, that when there is no query run that we dont display the useles 0ms part, so it will be just

    Sql Log ( 0 )

markstory:
Looks good but tests are failing.

dereuromark:
1) DebugKit\Test\TestCase\Panel\SqlLogPanelTest::testSummary
    Failed asserting that '0' matches PCRE pattern "/\d+ \/ \d+ ms/".

Any idea why that `$articles->findById(1)->first()` call doesnt count as query?

markstory:
@dereuromark That is pretty odd.

dereuromark:
I am hesitent to just adjust the assert.. :)

markstory:
Tests fixed and merged.

Issue #354, opened 2015-06-29
VariablesPanel crashes with Unbuffered queries


snelg:
If a Query is unbuffered, then after one call to toArray() any subsequent calls to toArray() will throw an Exception in ResultSet->rewind.
Example:

*index.ctp*
```php
foreach($clubs as $club) {
    pr($club);
}
```

*ClubsController.php*
```php
$query = $clubs->find();
$query->bufferResults(false);
$this->set('clubs', $query);
```

Without that "bufferResults(false)" call, everything's fine. But with that call, the page throws an exception.

So although the change in version 3.1.4 to show Query items instead of debug info is usually useful, there needs to be a fallback for this "unbuffered" condition.

Explicitly checking whether the query is buffered might be more proper, but I figured that we probably don't want DebugKit to crash the page if there's any sort of database exception while trying to populate this second copy of the view vars. So the pull request I'm submitting simply wraps a try/catch around the toArray function and falls back to __debugInfo calls.

markstory:
Closing as a fix has been merged.

PR #355, opened 2015-06-29
Issue #354 fix - Fallback for unbuffered queries in VariablesPanel

snelg:
Simple try/catch wrapper added as a fallback for unbuffered Query objects in VariablesPanel. Fixes issue #354


markstory:
Thanks :+1:

Issue #349, opened 2015-06-15
Simpler DebugKit Sql query out


markstory:
_From @cmann1 on June 11, 2015 13:2_

Perhaps an option to output simpler queries could be helpful when debugging, currently if you open `Sql Log` tab your queries will be extremely verbose:
```sql
SELECT Languages.id AS `Languages__id`, Languages.name AS `Languages__name`, Languages.code AS `Languages__code`, Languages.iso_code AS `Languages__iso_code`, Languages.is_default AS `Languages__is_default`, Languages.deleted AS `Languages__deleted` FROM languages Languages WHERE deleted = 0 ORDER BY Languages.id desc LIMIT 20 OFFSET 0
```
All of the `Table.col AS Table__col` can be difficult to look through and, in my experience at least, most of the time is not needed.
Maybe the debug kit could format the query, stripping out unnecessary information such as the default `AS` statements.
The above statement would be a lot easier to read and still provide the same information if the `SELECT` part was just: `Languages.id, Languages.name, ...`

If you need to see the original full query maybe it could be shown when you hover over it?

_Copied from original issue: cakephp/cakephp#6784_

markstory:
I don't really know if that is a good idea. SQL is non-trivial to parse and manipulate, and logging different SQL than was actually executed seems like it could cause as many problems as it solves.

markstory:
_From @ndm2 on June 11, 2015 20:28_

I'd rather like to optionally see the queries in a highlighted, maybe even formatted fashion, or an easier way too hook into the output generation.

For now I've created a plugin with a custom panel extending `SqlLogPanel`, which grabs the `DebugLog` logger, and uses a copy of the `sql_log_panel.ctp` element, where the query output is being formatted using https://github.com/jdorn/sql-formatter.

While this works, it would be nice if something similar would be supported out of the box, or easier to achieve, ie without having to "reimplement" too much code, namely the element in this case.

markstory:
Formatting the queries sounds like a good option.

markstory:
I gave using jdorn/sql-formatter a spin. The end result looks like:

![screen shot 2015-07-25 at 22 10 05](https://cloud.githubusercontent.com/assets/24086/8892192/083e90e0-331a-11e5-8d69-973018ad4b46.png)

The syntax highlighting uses garish colours which don't look good with the rest of DebugKit.

ndm2:
The colors can easily be customized, static properties aren't that nice tough.

[...]

PR #358, opened 2015-07-26
Format SQL queries so they are easier to read.

markstory:
Use a library to format SQL queries into a more readable formatting.

:notebook: This does not include syntax highlighting.

Refs #349

### DIFF
--- a/src/Template/Element/sql_log_panel.ctp
+++ b/src/Template/Element/sql_log_panel.ctp
@@ -16,6 +16,13 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 $noOutput = true;
+
+// Configure sqlformatter colours.
+SqlFormatter::$quote_attributes = 'style="color: #004d40;"';
+SqlFormatter::$backtick_quote_attributes = 'style="color: #26a69a;"';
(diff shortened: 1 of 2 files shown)


AD7six:
Is it worth making that configurable, i.e. so the user can do nothing/whitespace/syntax highlighting as they see fit?

src/Template/Element/sql_log_panel.ctp
```diff
@@ -63,7 +63,7 @@ $noOutput = true;
             <tbody>
                 <?php foreach ($queries as $query): ?>
                 <tr>
-                    <td><?= h($query['query']) ?></td>
+                    <td><pre><?= SqlFormatter::format($query['query'], false) ?></pre></td>
```
ndm2: This should stil use `h()`, as entity encoding is only being done when using syntax highlighting.

markstory: OK. Given that @lorenzo thinks adding syntax highlighting is important, I'll be adding that tonight.

markstory:
Queries have colours now.

lorenzo:
Fantastic! :clap:

Issue #359, opened 2015-07-27
Cache::config() and Cache::groupConfigs()


saraer:
I noticed config() and groupConfigs() methods of Cache(via DebugEngine) do not return correct value.
```php
debug(Cache::config('default'));
```
output
```php
'className' => object(DebugKit\Cache\Engine\DebugEngine) {
// this is not config array
    [protected] _config => [
        // here
}
```

markstory:
Pull request up now.

PR #360, opened 2015-07-30
Make config() proxy more seamlessly.

markstory:
This makes the DebugEngine appear as if it is not there which is nice.

Refs #359


ravage84:
What does "confuse" mean?

PR #361, opened 2015-07-30
add php7 build

antograssiot:
and remove deprecated option --dev for composer


ravage84:
Not sure about the impact of ``--prefer-dist`` in particular but :+1:

lorenzo:
Thanks!

PR #378, opened 2015-09-29
Moved routes to config/routes.php

jeremyharris:
This little quirk caused a bit of pain lately when I was reloading routes and debug kit wen't into an infinite XHR loop. This allows us to reload the plugin's routes after `Router::reload()` without having to reload all bootstrap setup. The only current downside I can see to loading bootstrap again when needing the routes is that the dispatcher filter is added again. But, in the future, any additional changes to bootstrap could cause problems.

This is more of an RFC I think, because it will cause "breaking" changes in existing applications, which could be more headache than it's worth. The app skel should be updated as well as the docs.

Thoughts on this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ php composer.phar require --dev cakephp/debug_kit "~3.0"
 
 * [Load the plugin](http://book.cakephp.org/3.0/en/plugins.html#loading-a-plugin)
 ```php
-Plugin::load('DebugKit', ['bootstrap' => true]);
+Plugin::load('DebugKit', ['bootstrap' => true, 'routes' => true]);
 ```
 * Set `'debug' => true,` in `config/app.php`.
 
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -15,7 +15,6 @@ use Cake\Datasource\ConnectionManager;
 use Cake\Event\EventManager;
 use Cake\Log\Log;
 use Cake\Routing\DispatcherFactory;
-use Cake\Routing\Router;
 use DebugKit\Routing\Filter\DebugBarFilter;
 
 $debugBar = new DebugBarFilter(EventManager::instance(), (array)Configure::read('DebugKit'));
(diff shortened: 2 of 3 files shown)


ADmad:
IMO breaking applications in development mode is acceptable in next minor release of debukit, given that debugkit is a development tool and not meant to be used in production.

jeremyharris:
Shall I go ahead and submit PRs to the docs and app skel? Or wait until this is merged if others agree?

ADmad:
Let others weight in :)

README.md
```diff
@@ -24,7 +24,7 @@ php composer.phar require --dev cakephp/debug_kit "~3.0"
 
 * [Load the plugin](http://book.cakephp.org/3.0/en/plugins.html#loading-a-plugin)
 ```php
-Plugin::load('DebugKit', ['bootstrap' => true]);
+Plugin::load('DebugKit', ['bootstrap' => true, 'routes' => true]);
```
markstory: This will break all the existing apps :cry: Is there a way we could move the routes that doesn't also break existing apps?

config/bootstrap.php
```diff
@@ -43,27 +42,6 @@
     ]);
 }
 
-Router::plugin('DebugKit', function ($routes) {
```
markstory: Here could we check to see if the plugin was loaded with routes, and conditionally require_once the routes file if they forgot?

README.md
jeremyharris: After looking into it, it doesn't seem possible as-is because `Plugin::routes()` requires `'routes' => true`, so trying to load them individually later doesn't work, making this PR moot since it would still require running `Plugin::bootstrap('DebugKit')` in order to load the routes.

config/bootstrap.php
jeremyharris: I think my above comment addresses this as well. Unfortunately it doesn't seem possible right now :( I'm not stoked on the idea of breaking everyone's debug env, but any `Router::reload()` in a debug env currently breaks debug kit. The `Plugin::bootstrap('DebugKit')` workaround I've been using *works* but I foresee all sorts of potential problems.

markstory: Why are you reloading the Router?

jeremyharris: I'm splitting functionality up into plugins. These plugins check for a sub domain, and create new routes based on that domain and allow certain access to the app based on the sub domain. I tried just overwriting the scope but that didn't work.

Perhaps this isn't the best approach and I'm sure my use case isn't a popular one. I didn't have anyone to consult with :P

markstory: Scopes can't really be overwritten, only appended into.

Have you tried using `Plugin::routes('DebugKit')` to see if routes were enabled? If they are not they can be conditionally loaded.

jeremyharris: Not sure what you mean. `Plugin::routes('DebugKit')` won't even try loading routes if the plugin isn't loaded with `'routes' => true`. (see: https://github.com/cakephp/cakephp/blob/master/src/Core/Plugin.php#L335)

markstory: Right, but that return value lets you know whether or not the plugin was loaded with routes. If it wasn't you can load them here.

jeremyharris: Alright I'll look into it and hopefully have something better tomorrow.

jeremyharris:
Added backwards compatibility.

markstory:
Nice work @jeremyharris

jeremyharris:
@markstory np. If you happen to remember, feel free to ping me when this is in a release and I'll submit PRs to the docs and app skeleton suggesting the new practice.

Issue #380, opened 2015-10-06
Database gets silly big


dakota:
If you've got a particularly active development environment (Especially if you are using simple ajax polling) it is possible for the sqlite db file to get massive:

```
vagrant@dev2:/var/www/html$ ls -hlaG tmp | grep sql
-rwxrwxrwx 1 vagrant 2.0G Oct  6 08:57 debug_kit.sqlite
```

While there is the `gc` method, I don't think that is is strict enough. It doesn't make sense to only delete requests older than 2 weeks when DebugKit only shows the last 10 requests at any one time. I believe that the `gc` should rather delete all requests, except for the last say 50 (configurable perhaps).

dereuromark:
:+1: No point in keeping anything beyond 20+
You wouldn't even know what it was about either way after xx other requests following.

ravage84:
:+1: to be configurable, those who actually need more shouldn't be bitten in the a**.
But I would say 50 sound good, max 100 may be.

How much space could be taken up by 100 requests?

dakota:
Closing this, I've opened a PR.

PR #381, opened 2015-10-06
Only keep the last x amount of requests

dakota:
Refs #380

### DIFF
--- a/src/Model/Table/RequestsTable.php
+++ b/src/Model/Table/RequestsTable.php
@@ -12,6 +12,7 @@
  */
 namespace DebugKit\Model\Table;
 
+use Cake\Core\Configure;
 use Cake\ORM\Query;
 use Cake\ORM\Table;
 use DebugKit\Model\Table\LazyTableTrait;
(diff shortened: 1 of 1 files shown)


src/Model/Table/RequestsTable.php
```diff
@@ -79,22 +80,23 @@ public function gc()
         if (time() % 100 !== 0) {
             return;
         }
-        $purge = $this->find()
+        $noPurge = $this->find()
             ->select(['id'])
             ->hydrate(false)
-            ->where(['requested_at <=' => new \DateTime('-2 weeks')])
+            ->order(['requested_at' => 'desc'])
+            ->limit(Configure::read('DebugKit.requestCount') ?: 20)
```
ravage84: Does it hurt (in size or performance) to keep a little bit more, e.g. 50 or 100, around?

With a little more request history, it will enable people to analyze the logs without having to configure a higher count and then go and build up the history first.

dakota: As it stands the history panel will only ever show at most 10 requests, so unless you would like to delve into the serialized mess that is the raw db I don't think it makes sense to keep more than 20 (At least by default).

ravage84: True, but then it is even pointless to keep more than 10, as it stands.

On the other hand, is there a way to change how many requests are shown?

Or how about a "Show all" button to show all kept entries in the request history?
This way it makes sense again to keep more than it ormally displays...

markstory:
I didn't consider polling scenarios when I added the initial gc code. Keeping 20 by default seems pretty reasonable.

PR #382, opened 2015-10-06
Removed inline javascript and style code to support Content Security Policy

TheFRedFox:
Based on the issue [#7511](https://github.com/cakephp/cakephp/issues/7511) in cakephp repo, I removed inline javascript and style code and replace them with code in separated js resp. css files.
So DebugKit is (hopefully) fully supported for Content Security Policy. DebugKit also set its own Content-Security-Policy, so it should be guaranteed that everything can be loaded by DebugKit.

### DIFF
--- a/src/Controller/RequestsController.php
+++ b/src/Controller/RequestsController.php
@@ -36,6 +36,8 @@ class RequestsController extends Controller
         if (!Configure::read('debug')) {
             throw new NotFoundException();
         }
+
+        $this->response->header(['Content-Security-Policy' => "default-src 'self'; style-src 'self' fonts.googleapis.com; font-src 'self' fonts.gstatic.com"]);
     }
 
     /**
--- a/src/Controller/RequestsController.php
+++ b/src/Controller/RequestsController.php
@@ -37,7 +37,7 @@ class RequestsController extends Controller
             throw new NotFoundException();
         }
 
-        $this->response->header(['Content-Security-Policy' => "default-src 'self'; style-src 'self' fonts.googleapis.com; font-src 'self' fonts.gstatic.com"]);
+        $this->response->header(['Content-Security-Policy' => '']);
     }
 
     /**
--- a/src/Routing/Filter/DebugBarFilter.php
+++ b/src/Routing/Filter/DebugBarFilter.php
@@ -234,7 +234,6 @@ class DebugBarFilter extends DispatcherFilter
         }
         $url = Router::url('/', true);
         $script = "<script id=\"__debug_kit\" data-id=\"{$id}\" data-url=\"{$url}\" src=\"" . Router::url('/debug_kit/js/toolbar.js') . '"></script>';
-        $script .= '<link rel="stylesheet" href="' . Router::url('/debug_kit/css/iframe.css') . '"/>';
         $body = substr($body, 0, $pos) . $script . substr($body, $pos);
         $response->body($body);
     }
--- a/src/Routing/Filter/DebugBarFilter.php
+++ b/src/Routing/Filter/DebugBarFilter.php
@@ -233,7 +233,6 @@ class DebugBarFilter extends DispatcherFilter
             return;
         }
         $url = Router::url('/', true);
-        //$script = "<script>var __debug_kit_id = '${id}', __debug_kit_base_url = '${url}';</script>";
         $script = "<script id=\"__debug_kit\" data-id=\"{$id}\" data-url=\"{$url}\" src=\"" . Router::url('/debug_kit/js/toolbar.js') . '"></script>';
         $script .= '<link rel="stylesheet" href="' . Router::url('/debug_kit/css/iframe.css') . '"/>';
         $body = substr($body, 0, $pos) . $script . substr($body, $pos);
--- a/src/Routing/Filter/DebugBarFilter.php
+++ b/src/Routing/Filter/DebugBarFilter.php
@@ -233,8 +233,9 @@ class DebugBarFilter extends DispatcherFilter
             return;
         }
         $url = Router::url('/', true);
-        $script = "<script>var __debug_kit_id = '${id}', __debug_kit_base_url = '${url}';</script>";
-        $script .= '<script src="' . Router::url('/debug_kit/js/toolbar.js') . '"></script>';
+        //$script = "<script>var __debug_kit_id = '${id}', __debug_kit_base_url = '${url}';</script>";
+        $script = "<script id=\"__debug_kit\" data-id=\"{$id}\" data-url=\"{$url}\" src=\"" . Router::url('/debug_kit/js/toolbar.js') . '"></script>';
+        $script .= '<link rel="stylesheet" href="' . Router::url('/debug_kit/css/iframe.css') . '"/>';
         $body = substr($body, 0, $pos) . $script . substr($body, $pos);
         $response->body($body);
     }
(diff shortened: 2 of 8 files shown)


webroot/js/debug_kit.js
```diff
@@ -0,0 +1,24 @@
...
+
+$(document).ready(function() {
+    toolbar = new Toolbar({
+        button: $('#toolbar'),
+        content: $('#panel-content-container'),
+        panelButtons: $('.panel'),
+        panelClose: $('#panel-close'),
+        keyboardScope : $(document),
+        currentRequest: '<?= $toolbar->id ?>',
+        originalRequest: '<?= $toolbar->id ?>',
```
markstory: These PHP tags look wrong.

webroot/js/toolbar.js
```diff
@@ -29,7 +37,13 @@
...
 		var body = doc.body;
 		iframe = doc.createElement('iframe');
-		iframe.setAttribute('style', 'position: fixed; bottom: 0; right: 0; border: 0; outline: 0; overflow: hidden; z-index: 99999;');
+        iframe.style.position = 'fixed';
+        iframe.style.bottom = 0;
+        iframe.style.right = 0;
+        iframe.style.border = 0;
+        iframe.style.outline = 0;
+        iframe.style.overflow = 'hidden';
+        iframe.style.zIndex = 99999;
```
markstory: Why did this have to change?

markstory: Because CSP nerfs style attribute evaluation.

dakota: CSP seems to break more things than it fixes...

src/Controller/RequestsController.php
```diff
@@ -36,6 +36,8 @@ public function beforeFilter(Event $event)
         if (!Configure::read('debug')) {
             throw new NotFoundException();
         }
+
+        $this->response->header(['Content-Security-Policy' => "default-src 'self'; style-src 'self' fonts.googleapis.com; font-src 'self' fonts.gstatic.com"]);
```
Marlinc: Would it make sense to instead of using ```'self'``` use the direct path to the assets?

webroot/js/debug_kit.js
Marlinc: This probably has to be ```elem.getAttribute("data-id");```.

webroot/js/toolbar.js
TheFRedFox: Yeah exactly @markstory . I tried it with setAttribute, but it won't work. You can just change styles with directly accessing the properties separated with e.g. ```iframe.style.position``` .

TheFRedFox: @dakota Well I won't say that CSP is breaking things, but you have to do some changes, which could be a bit tough work to do. But it will bring much protection against XSS attacks if configured right. :wink: 
For example I also won't say 64 bit support for a program will break more things than it fixes, but you have to do much work to change the program to support 64 bit. :wink:

src/Controller/RequestsController.php
TheFRedFox: I didn't see an option to set a direct asset path, but I shouldn't be necessary, because everything on your own server should be ok to load. If the assets on your server are attacked .. well .. then you have more problems than just XSS attacks. :smile: 
I will look if there is an option to a direct asset path.

webroot/js/debug_kit.js
TheFRedFox: Oh yeah, holy ... :smile: Interesting that it is working nevertheless. ^^ I will fix it.

markstory:
So I noticed a few regressions that will need to be figured out before this can be merged:

* Syntax highlighting in the SQL panel uses inline styles and is currently broken.
* Waterfall chart in the Timer panel uses inline styles and is currently broken.

Other than these issues the changes are looking good.

TheFRedFox:
@markstory I fixed Syntax highlighting as we can just separate all styles in a css file and load the css file in the ```sql_log_panel.ctp``` directly.
BUT
The timer panel give me headache! As we can't just load the script file in the ```timer_panel.ctp``` directly because of ```eval``` restrictions of CSP. To just separate it in an existing js file is possible, but it will definitely completely restrict js usage for new user's panels.

To be honest ... as it is in an own iframe with it's own CSP restrictions and as it is just supposed to be used in development ... I would suggest just to allow ```inline script```, ```inline style``` and ```inline eval```, so everything can be load in DebugKit. (or maybe we can just force allow everything for CSP for DebugKit :wink:)

src/Routing/Filter/DebugBarFilter.php
```diff
@@ -233,8 +233,7 @@ protected function _injectScripts($id, $response)
             return;
         }
         $url = Router::url('/', true);
-        $script = "<script>var __debug_kit_id = '${id}', __debug_kit_base_url = '${url}';</script>";
-        $script .= '<script src="' . Router::url('/debug_kit/js/toolbar.js') . '"></script>';
+        $script = "<script id=\"__debug_kit\" data-id=\"{$id}\" data-url=\"{$url}\" src=\"" . Router::url('/debug_kit/js/toolbar.js') . '"></script>';
```
TheFRedFox: This is absolutely needed for CSP support inside the original page.

webroot/js/toolbar.js
```diff
@@ -1,3 +1,11 @@
+var __debug_kit_id, __debug_kit_base_url;
+var elem = document.getElementById("__debug_kit");
+if (elem) {
+    __debug_kit_id = elem.getAttribute("data-id");
+    __debug_kit_base_url = elem.getAttribute("data-url");
+    elem = null;
+}
+
```
TheFRedFox: This is absolutely needed for CSP support inside the original page.

webroot/js/toolbar.js
TheFRedFox: This is absolutely needed for CSP support inside the original page.

tests/TestCase/Routing/Filter/DebugBarFilterTest.php
```diff
@@ -146,9 +146,8 @@ public function testAfterDispatchSavesData()
         $this->assertEquals('Sql Log', $result->panels[7]->title);
 
         $expected = '<html><title>test</title><body><p>some text</p>' .
-            "<script>var __debug_kit_id = '" . $result->id . "', " .
-            "__debug_kit_base_url = 'http://localhost/';</script>" .
-            '<script src="/debug_kit/js/toolbar.js"></script>' .
+            '<script id="__debug_kit" data-id="' . $result->id . '" ' .
+            'data-url="http://localhost/" src="/debug_kit/js/toolbar.js"></script>' .
```
TheFRedFox: This is absolutely needed for CSP support inside the original page.

src/Controller/RequestsController.php
markstory: Given that the toolbar is no longer requiring csp, this should probably be removed.

markstory:
I think the current state where debug kit can be loaded into a csp page, but does not use csp itself is a good approach to take. :+1:

src/Controller/RequestsController.php
TheFRedFox: Removed or replaced with ```null```? So it will overwrite possible header options which were set before somewhere.

TheFRedFox:
Ok .. I just changed it so the CSP header will be force unset for the possibility that the user set the CSP header somewhere else in the original application (e.g. directly in bootstrap).
It should work now with everything inside the iframe. ;)

PR #384, opened 2015-10-21
Correctly set the connection when using fixtures.

markstory:
When using fixtures outside of the test harness, correctly set the active connection. This should help future proof any future connection related features that are introduced into fixtures.

Refs cakephp/cakephp#7586

### DIFF
--- a/src/Model/Table/LazyTableTrait.php
+++ b/src/Model/Table/LazyTableTrait.php
@@ -45,7 +45,7 @@ trait LazyTableTrait
             if ($class === false) {
                 throw new \RuntimeException("Unknown fixture '$name'.");
             }
-            $fixture = new $class();
+            $fixture = new $class($this->connection()->configName());
             $table = $fixture->table;
             if (in_array($table, $existing)) {
                 continue;
--- a/tests/Fixture/PanelsFixture.php
+++ b/tests/Fixture/PanelsFixture.php
@@ -21,7 +21,6 @@ use Cake\TestSuite\Fixture\TestFixture;
  */
 class PanelsFixture extends TestFixture
 {
-
     /**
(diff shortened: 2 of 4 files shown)


src/Model/Table/LazyTableTrait.php
```diff
@@ -45,7 +45,7 @@ public function ensureTables(array $fixtures)
             if ($class === false) {
                 throw new \RuntimeException("Unknown fixture '$name'.");
             }
-            $fixture = new $class();
+            $fixture = new $class($this->connection()->configName());
```
ADmad: I tried this fix and it doesn't seem to help. [TestFixture::__construct()](https://github.com/cakephp/cakephp/blob/master/src/TestSuite/Fixture/TestFixture.php#L95) doesn't take any arguments.

HavokInspiration: This might indeed need a modification of the constructor of TestFixture. There is currently no way to set the connection to use by a Fixture except in the property field (and the change to the fixtures of the plugin might cause strict notices).

HavokInspiration:
Another idea could be to switch to using migrations using the Migrations class. But that would add the migrations plugin as a dependency.
Might not be the best idea but I thought it would be worth mentionning it.

src/Model/Table/LazyTableTrait.php
markstory: The fixtures in DebugKit had constructors added.

HavokInspiration:
I tested the fix on a fresh install and it fixed the issue for me.

markstory:
Thanks for doing the release @lorenzo

ADmad:
:+1:

PR #398, opened 2015-12-30
Add summary count to Include Panel.

markstory:
Add the total number of files included in the include panel. This makes 'fat' requests a bit more visible.

### DIFF
--- a/tests/TestCase/Panel/IncludePanelTest.php
+++ b/tests/TestCase/Panel/IncludePanelTest.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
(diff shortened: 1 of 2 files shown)


lorenzo:
Nice!

tests/TestCase/Panel/IncludePanelTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Test that the log panel outputs a summary.
+     *
+     * @return void
+     */
+    public function testSummary()
+    {
+        $total = $this->panel->summary();
+        $this->assertGreaterThan(50, $total);
```
lorenzo: It will be fun to see how this assertion breaks as we change cake :P

markstory: If we get under 50 I'll be :tada:

Issue #392, opened 2015-12-24
Unserialize errors after moving to PostgreSQL storage


ethanpooley:
I'm running into problems making PostgreSQL the storage mechanism for DebugKit. Features that worked previously (History, SQL Log, and Variables) now print `unserialize()` errors.

Versions:
- Cake 3.1.3
- DebugKit 3.2.1
- PHP 5.6.16
- PostgreSQL 9.4.5

Environments:
- Ubuntu 14.04 on Virtualbox 5.0.12
- Ubuntu 14.04 on Heroku (Cedar-14)

Configuration:
- `config/app.php`
```
    'Datasources' => [
        'debug_kit' => [
            'className' => 'Cake\Database\Connection',
            'driver' => 'Cake\Database\Driver\Postgres',
            'persistent' => false,
            'host' => 'localhost',
            'port' => '5432',
            'username' => 'REDACTED',
            'password' => 'REDACTED',
            'database' => 'REDACTED',
            'encoding' => 'utf8',
            'timezone' => 'UTC',
            'cacheMetadata' => true,
            'quoteIdentifiers' => false,
        ],
        ...
```
- `config/bootstrap.php`
```
if (Configure::read('debug')) {
    Plugin::load('DebugKit', ['bootstrap' => true, 'routes' => true]);
}
```

Output when clicking on the Variables panel:
```
Notice (8): unserialize(): Error at offset 129 of 133 bytes [ROOT/vendor/cakephp/debug_kit/src/Controller/PanelsController.php, line 91]
Code

        $panel = $this->Panels->get($id);
        $this->set('panel', $panel);
        $this->set(unserialize($panel->content));

Context

$id = 'd8e536ac-cf60-4607-96c6-917d7bf52f1f'
$panel = object(DebugKit\Model\Entity\Panel) {

	'id' => 'd8e536ac-cf60-4607-96c6-917d7bf52f1f',
	'request_id' => 'ceacdb37-3521-48ce-b28b-911c825c77bd',
	'panel' => 'Variables',
	'title' => 'Variables',
	'element' => 'DebugKit.variables_panel',
	'summary' => '3',
	'content' => 'a:2:{s:7:"content";a:3:{s:10:"_serialize";a:1:{i:0;s:8:"agencies";}s:8:"agencies";a:20:{i:0;O:23:"App\Model\Entity\Agency":10:{s:14:"',
	'[new]' => false,
	'[accessible]' => [
		'*' => true
	],
	'[dirty]' => [],
[...]

PR #400, opened 2016-01-05
Make panel data binary values.

markstory:
postgres gets really sad when you put null bytes into character columns. With this change new schema will use binary columns which work fine in postgres.

The entity accessor makes reading small and big blobs work the same.

Refs #392

### DIFF
--- a/src/Model/Entity/Panel.php
+++ b/src/Model/Entity/Panel.php
@@ -25,4 +25,20 @@ class Panel extends Entity
      * @var array
      */
     protected $_hidden = ['content'];
+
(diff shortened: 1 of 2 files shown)


lorenzo:
Nice, good you could find the cause :D

ethanpooley:
Oh good grief. I've literally quoted the PHP manual on that on SO before (https://stackoverflow.com/questions/10770974/which-mysql-column-type-for-serializedata/27389223#27389223) and I still didn't think of it here. Nice find @markstory. Thank you!

PR #403, opened 2016-02-09
Forcing the panel to use the default cake's View class

lorenzo:
When having an altered AppView class, the toolbar can break


markstory:
I'll do a release tonight for this.

PR #412, opened 2016-03-16
In certain circumstances, when beforeRender is invoked multiple times…

theplankmeister:
…the debugToolbarPanels variable ends up recursively containing itself. Hundreds of thousands of deeply nested HTML tags, along with the JS that traverses them, make browsers very sad.


lorenzo:
Thanks

Issue #416, opened 2016-04-14
Include a standalone AjaxView


markstory:
Leaning on the AjaxView that come bundled in the Application is problematic becuase developers can change either the AjaxView or its parent classes in ways that are not compatible with DebugKit. Having a standalone view class will allow debugkit to exist in hostile applications. See cakephp/app#366.

markstory:
Pull request open now.

PR #417, opened 2016-04-14
Add a standalone AjaxView

markstory:
Having a self-contained view class will allow debugkit to continue working even when the application doesn't have an AjaxView, or when that AjaxView/AppView is not suitable for rendering debugkit.

Refs #416


cpierce:
👍

src/View/AjaxView.php
```diff
@@ -0,0 +1,36 @@
...
+ * @since         3.2.8
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace DebugKit\View;
+
+use Cake\View\View;
+
+/**
+ * A view class that is used for AJAX responses.
+ * Currently only switches the default layout and sets the response type -
```
ravage84: New line

PR #421, opened 2016-05-27
Fix history loading and sorting issues

markstory:
The history panel was relying on the recently fixed RequestHandler bug, and sorting has been broken for a while.


lorenzo:
Good job!

PR #427, opened 2016-07-13
Make word-break property compatible with Firefox Browser

albertcansado:
Firefox not implement break-word value but accepts break-all with similar effect

Now (without break-all) on Firefox

captura de pantalla 2016-07-13 a las 16 44 39

### DIFF
--- a/webroot/css/toolbar.css
+++ b/webroot/css/toolbar.css
@@ -66,7 +66,7 @@ p {
 }
 
 .hidden {
-    display: none;
+	display: none;
 }
 
 /* Open close button */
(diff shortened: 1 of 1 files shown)


webroot/css/toolbar.css
```diff
@@ -232,6 +232,7 @@ table tbody tr:nth-child(odd) {
 }
 .debug-table td:nth-child(n+2) {
 	word-wrap: break-word;
+    word-break: break-all;
```
ADmad: Indentation needs to be fixed.

markstory:
Thanks!

PR #436, opened 2016-08-31
Don't load DebugKit in the phpdbg cli

jeremyharris:
Following Cake's support of `phpdbg`, DebugKit shouldn't load in the `phpdbg` command line environment either.


markstory:
Thanks :+1:

PR #437, opened 2016-09-01
Add missing direct dependency

dereuromark:
Refs https://github.com/cakephp/bake/issues/268


lorenzo:
As a note, the dependency was removed in the past because it caused pin to maintain when bumping major versions of cakephp.

dereuromark:
The Asterix Wildcard should allow for any version here.

PR #439, opened 2016-09-06
Add Explain button to SQL log panel

chinpei215:
After cakephp/cakephp#9421 merged, DebugKit will show you Explain buttons in SQL log panel. I confirmed working with Vivaldi 1.3, IE 11, Edge 25, Chrome 53 and FF 48, on WIndows.

### MySQL
explain-mysql

### PostgreSQL
explain-postgres

### SQLite
explain-sqlite

### DIFF
--- a/src/Controller/ToolbarController.php
+++ b/src/Controller/ToolbarController.php
@@ -15,8 +15,11 @@ namespace DebugKit\Controller;
 use Cake\Cache\Cache;
 use Cake\Controller\Controller;
 use Cake\Core\Configure;
+use Cake\Datasource\ConnectionManager;
 use Cake\Event\Event;
+use Cake\Network\Exception\BadRequestException;
 use Cake\Network\Exception\NotFoundException;
+use Cake\Utility\Security;
 
 /**
  * Provides utility features need by the toolbar.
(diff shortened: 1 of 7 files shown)


src/Controller/ToolbarController.php
```diff
@@ -71,4 +74,57 @@ public function clearCache()
...
+        $data = $this->request->data('data');
+        $hash = $this->request->data('hash');
+
+        if (Security::hash($data, null, true) !== $hash) {
+            throw new BadRequestException('Invalid hash');
+        }
+
+        $data = json_decode($data, true);
+        if (!$data) {
+            throw new BadRequestException('Invalid json');
```
dereuromark: JSON

dereuromark:
Nice :)

src/Template/Element/sql_log_panel.ctp
```diff
@@ -65,14 +71,38 @@ SqlFormatter::$pre_attributes = 'style="color: #222; background-color: transpare
...
+                        $data = json_encode([
+                            'connection' => $logger->name(),
+                            'query' => $query['queryString'],
+                            'params' => $query['params']
+                        ]);
+                        echo $this->Form->create();
+                        echo $this->Form->hidden('data', ['value' => $data]);
+                        echo $this->Form->hidden('hash', ['value' => Security::hash($data, null, true)]);
+                        echo $this->Form->button('Explain', ['type' => 'button', 'class' => 'btn-primary sql-explain-link']);
+                        echo $this->Form->end();
```
markstory: Is this something we could use postLink() or postButton() for?

src/Template/Element/sql_log_panel.ctp
```diff
@@ -84,3 +114,57 @@ SqlFormatter::$pre_attributes = 'style="color: #222; background-color: transpare
...
+<script>
+$(document).ready(function() {
+    var baseUrl = '<?= $this->Url->build([
+        'plugin' => 'DebugKit',
+        'controller' => 'Toolbar',
+        'action' => 'sqlExplain'
+    ]); ?>';
+
+    function createTable(result) {
+        var $table = $('<table>');
```
markstory: I don't think the `$` prefixes are necessary here.

markstory:
Should we update the composer.json to require the new version of cakephp/cakephp? That might remove the need for `method_exists` checks.

ravage84:
@markstory might be an option, but I wouldn't do that in a patch release then...

markstory:
@ravage84 I was going to make the explain button go into a minor release of DebugKit. Its a pretty big new feature.

chinpei215:
Thank you for your comments.

> Should we update the composer.json to require the new version of cakephp/cakephp? That might remove the need for method_exists checks.

I am fine to target next minor release. Then we can remove `isset($query->queryString)` check from DebugLog. However, we would still need to call `method_exists`, as the canExplain method is not contained in ConnectionInterface.

src/Template/Element/sql_log_panel.ctp
chinpei215: Ah sorry. It's my personal habit 😛 When I assign an instance of jQuery object assign to a variable, I always prefix `$` to the name. I will remove all of them.

PR #440, opened 2016-09-06
Fixing a notice when a PSR logger is used

burzum:


src/Database/Log/DebugLog.php
```diff
@@ -122,7 +122,11 @@ public function totalRows()
     public function log(LoggedQuery $query)
     {
         if ($this->_logger) {
-            $this->_logger->log($query);
+            if ($this->_logger instanceof \Psr\Log\AbstractLogger) {
```
dereuromark: I would put this FQCN in use statement.

burzum: Well, if we do so I would do this:

```php
use \Psr\Log\AbstractLogger as PsrAbstractLogger;
```

markstory: `use \Psr\Log\AbstractLogger as PsrAbstractLogger;` works for me.

PR #441, opened 2016-09-08
Always setting layout and class

jeremyharris:
DebugKit seemed to be inheriting our view class and layout, which in turn was loading our app's CSS and JS and breaking the toolbar (and taking a long time as it had to load a ton of stuff).

I don't have a good understanding behind the bug caused by https://github.com/cakephp/debug_kit/commit/bcf8b1a8cd2b56d32985d6759e2874ef95f89336, so I want to be sure that reverting it won't cause people's panels to break.

Pinging @markstory as you're the guy who wrote the commit 😄

### DIFF
--- a/src/Controller/PanelsController.php
+++ b/src/Controller/PanelsController.php
@@ -53,9 +53,11 @@ class PanelsController extends Controller
      */
     public function beforeRender(Event $event)
     {
-        $this->viewBuilder()
-            ->layout('DebugKit.panel')
-            ->className('DebugKit.Ajax');
+        $this->viewBuilder()->layout('DebugKit.toolbar');
+
+        if (!$this->request->is('json')) {
+            $this->viewBuilder()->className('DebugKit.Ajax');
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jeremyharris:
This might just be a problem with our app code. Our templating situation is ... interesting ...

src/Controller/PanelsController.php
```diff
@@ -53,11 +53,9 @@ public function beforeFilter(Event $event)
...
     public function beforeRender(Event $event)
     {
-        $builder = $this->viewBuilder();
-        if (!$builder->className()) {
-            $builder->layout('DebugKit.panel')
-                ->className('DebugKit.Ajax');
-        }
+        $this->viewBuilder()
+            ->layout('DebugKit.panel')
+            ->className('DebugKit.Ajax');
```
markstory: This will break fetching the history data. I tested locally and it results in 500 errors containing `"Template file "Panels/index.ctp" is missing."`

jeremyharris: Ah, missed that. I'll look into it.

jeremyharris: @markstory sorry about the delay on this. I updated the builder to set the ajax view only if the request isn't JSON. As far as I can tell, only the PanelsController uses JSON.

PR #443, opened 2016-09-10
This seems to be needed with some cases of variable content.

dereuromark:
I got

>Cannot convert value to string - InvalidArgumentException

and DebugKit stopped execution completely.

Apperently it was about
```
[
	(int) 0 => '5c517388-a635-41e2-bc50-b691b0d659fc'
]
```
and Cake\Database\Type\StringType->toDatabase
ROOT\Vendor\cakephp\cakephp\src\Database\Type\UuidType.php, line 39

Shouldnt DebugKit in general catch more exceptions and try to not make it impossible to continue working by rather displaying empty content than going to exception case?

Also fixed the case for another exception.

### DIFF
--- a/src/Panel/VariablesPanel.php
+++ b/src/Panel/VariablesPanel.php
@@ -23,9 +23,10 @@ use Cake\Utility\Hash;
 use Closure;
 use DebugKit\DebugPanel;
 use Exception;
+use InvalidArgumentException;
 use PDO;
 use RuntimeException;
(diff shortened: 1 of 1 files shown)


markstory:
> Shouldnt DebugKit in general catch more exceptions and try to not make it impossible to continue working by rather displaying empty content than going to exception case?

Probably, but I'm not sure which exceptions people run into.

PR #446, opened 2016-09-21
various code fixes and phpdoc fixes

thinkingmedia:
I found a lot of little things missing and a few incorrectly named method called (maybe that part of the code isn't used anymore).

This PR contains a lot of changes. So please review it before merging.

Most of the changes are the result of syntax issues highlighted by phpstorm.

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -1,7 +1,8 @@
 <?php
+use Cake\Routing\RouteBuilder;
 use Cake\Routing\Router;
 
-Router::plugin('DebugKit', function ($routes) {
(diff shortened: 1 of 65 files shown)


thinkingmedia:
> Expected docblock to be aligned with code.

Does anyone know what this even means? I thought it was aligned!

ndm2:
Multiline docblocks are for member and class descriptions, inline hints should use single line comments like

```
/* @var $foo Bar */
```

Note the use of a single asterik only, while most IDEs recognize the two asterik variant too, CS require the former.

dereuromark:
Nice Job!

PR #448, opened 2016-09-22
Adds a base class for controllers

thinkingmedia:
Noticed last time that the controllers to be reduced a little if they used a common base class.

### DIFF
--- a/src/Controller/RequestsController.php
+++ b/src/Controller/RequestsController.php
@@ -37,6 +37,15 @@ class RequestsController extends DebugKitController
     }
 
     /**
+     * Do not load components for this controller.
+     *
+     * @return void
+     */
+    public function initialize()
+    {
+    }
+
+    /**
      * Before render handler.
      *
      * @param \Cake\Event\Event $event The event.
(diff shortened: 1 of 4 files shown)


thinkingmedia:
So apparently the class identifier is not supported in PHP 5.4.

Isn't 5.4 below the required PHP version for Cake 3?

ADmad:
> Isn't 5.4 below the required PHP version for Cake 3?

Cake 3.2 and below support 5.4 and hence Debugkit needs to work on PHP 5.4

thinkingmedia:
done.

src/Controller/RequestsController.php
```diff
@@ -34,12 +31,18 @@ class RequestsController extends Controller
...
+
+        parent::beforeFilter($event);
+    }
+
+    /**
+     * Do not load components for this controller.
+     *
+     * @return void
+     */
+    public function initialize()
```
markstory: Is the initialize state really 'common' if some of the controllers turn it off?

thinkingmedia: I thought it was, but that one controller has issues with the request handler. There are tests that say it's a JSON response but it has an HTML only view. Either I dig deep into the code to see why or revert the behavior for this one case.

thinkingmedia:
closing as not important.

PR #454, opened 2016-10-11
Improve local storage check for Safari private mode

nickygerritsen:
Safari in private browsing mode has the `localStorage` object available, but setting any item on it fails. See for example [this Stack Overflow post](http://stackoverflow.com/questions/14555347/html5-localstorage-error-with-safari-quota-exceeded-err-dom-exception-22-an).

This improved check should catch that error, thus allowing us to actually open de debug bar in Safari in a private window.

### DIFF
--- a/webroot/js/toolbar-app.js
+++ b/webroot/js/toolbar-app.js
@@ -49,14 +49,11 @@ Toolbar.prototype = {
 			if (!window.localStorage) {
 				this._localStorageAvailable = false;
 			} else {
-				try
-				{
+				try {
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Hmmm I adopted the indenting style of the rest of the file, which uses tabs. Should I change my indenting?

nickygerritsen:
It also seems travis fails with the SQLite database? I can't imagine that this is because of my commits.

lorenzo:
Thanks for the PR!

markstory:
I'll clean up the JS linting so its less spammy.

PR #460, opened 2016-10-19
filter credentials in EnvironmentPanel to hide username:password

steinkel:


justinatack:
This doesn't make sense. Why would you hide this information? The point of a debug panel is to have access to all information to "debug" problems.

markstory:
Not sure I get this either. Why do development server values need to be hidden?

steinkel:
The reason behind this PR was:
I received a screenshot of another user while trying to debug an issue in his dev environment, then I realized he was using his personal google email account (password displayed). 
I suggested this change to:
* Prevent accidental user/passwords shared
* For shared environments (like a QA environment shared with other devs), allow them to use debug_kit, but hide the credentials to the backend db/email/redis/others
* Minimize the information shared if debug enabled in public environments by mistake

For me it's a safer default, as the user/password does not get shared when configured in app.php (not using the environment).

I understand your point though, this would make debugging those env variables a bit harder as the username:password is hidden. So if you don't feel we should merge it I can move it to a Plugin instead.

Thank you!

markstory:
Would it be hard to do the filtering in the template? That would let us put tooltips with the unmasked value.

thinkingmedia:
You could easily share SQL that contains personal information, proprietary secrets, private keys and other  unexpected data.

The general sharing of anything from the debug_kit is a peek inside the black box of the server. While you've filtered this one use case there remains thousands more cases that aren't filtered.

Do not run the debug_kit on staging servers or production even if debug is enabled. If the feature has to be enabled than instruct your tests to copy/paste information that is needed, and to not share personal information.

If you do not want testers to see private keys, passwords, etc. Do not turn debug_kit on.

steinkel:
@markstory something like this would work?
(note tests missing, if you guys think this would be nice to have I'll fix that later)

Thanks,

webroot/css/toolbar.css
```diff
@@ -514,3 +514,8 @@ pre,
 		transform: rotate(360deg);
 	}
 }
+
+#panel-content .panel-content a.filtered-credentials {
```
markstory: You shouldn't need an ID selector, the panels are rendered in an iframe.

markstory:
Helper looks pretty good, could we get tests for it?

steinkel:
@markstory test added

Issue #457, opened 2016-10-12
Routes panel


andrej-griniuk:
https://github.com/andrej-griniuk/cakephp-debug-routes
Happy to contribute it to the core if you guys like it.
It super simple at the moment, does pretty much same thing `bin/cake routes` does, but any suggestions welcome.

markstory:
Sounds good to me.

markstory:
Closing as there is a routes panel now.

PR #461, opened 2016-10-25
Routes panel

andrej-griniuk:
Refs #457

### DIFF
--- a/src/Template/Element/routes_panel.ctp
+++ b/src/Template/Element/routes_panel.ctp
@@ -14,7 +14,7 @@
     </thead>
     <tbody>
     <?php foreach ($routes as $route): ?>
-        <tr<?php if ($matchedRoute === $route->template) echo ' class="highlighted"'; ?>>
+        <tr class="<?= ($matchedRoute === $route->template) ? 'highlighted' : '' ?>">
             <td><?= \Cake\Utility\Hash::get($route->options, '_name', $route->getName()) ?></td>
             <td><?= $route->template ?></td>
             <td><?= json_encode($route->defaults) ?></td>
--- a/src/Template/Element/routes_panel.ctp
+++ b/src/Template/Element/routes_panel.ctp
@@ -1,0 +1,24 @@
+<?php
+/**
+ * @var \Cake\Routing\Route\Route[] $routes
+ * @var string $matchedRoute
+ */
+?>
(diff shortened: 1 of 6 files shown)


src/Template/Element/routes_panel.ctp
```diff
@@ -0,0 +1,24 @@
...
+    <thead>
+    <tr>
+        <th><?= __d('debug_kit', 'Route name') ?></th>
+        <th><?= __d('debug_kit', 'URI template') ?></th>
+        <th><?= __d('debug_kit', 'Defaults') ?></th>
+    </tr>
+    </thead>
+    <tbody>
+    <?php foreach ($routes as $route): ?>
+        <tr<?php if ($matchedRoute === $route->template) echo ' class="highlighted"'; ?>>
```
dereuromark: Use <?= and ternary better for inliners.

src/Template/Element/routes_panel.ctp
```diff
@@ -0,0 +1,24 @@
...
+        <th><?= __d('debug_kit', 'Route name') ?></th>
+        <th><?= __d('debug_kit', 'URI template') ?></th>
+        <th><?= __d('debug_kit', 'Defaults') ?></th>
+    </tr>
+    </thead>
+    <tbody>
+    <?php foreach ($routes as $route): ?>
+        <tr class="<?= ($matchedRoute === $route->template) ? 'highlighted' : '' ?>">
+            <td><?= \Cake\Utility\Hash::get($route->options, '_name', $route->getName()) ?></td>
+            <td><?= $route->template ?></td>
```
markstory: Should HTML escape this to be safe.

src/Template/Element/routes_panel.ctp
```diff
@@ -0,0 +1,24 @@
...
+            <td><?= json_encode($route->defaults) ?></td>
```
markstory: Does `$this->Toolbar->makeNeatArray($route->defaults)` look OK here?

markstory: ok.

lorenzo:
Would it be better to fetch the routes dynamically in the panel instead of storing it in the database for each request? Routes are more or less static data, but they consume space and also time to serialize

andrej-griniuk:
Makes sense. Updated.

markstory:
Thanks!

Issue #471, opened 2017-01-13
Fatal error with DebugKit 3.5.1 and CakePHP 3.3.x


fquffio:
I just issued `composer update`, and I am getting this error:

```
Fatal error: Call to undefined method Cake\Network\Response::getBody() in MYPROJECT/vendor/cakephp/debug_kit/src/ToolbarService.php on line 213
```

I guess this is caused by a wrong requirement in `composer.json`: DebugKit 3.5.1 should require CakePHP ≥ 3.4.

_EDIT:_
My app has no special configuration: it just loads the `DebugKit` plugin as usual.
As a reference, here is my `composer info` output:
```
aura/installer-default                1.0.0              Installs Aura packages using the Composer defaults.
aura/intl                             1.1.1              The Aura.Intl package provides internationalization (I18N) tools, specifically
cakephp/bake                          1.2.11             Bake plugin for CakePHP 3.0
cakephp/cakephp                       3.3.11             The CakePHP framework
cakephp/cakephp-codesniffer           2.3.0              CakePHP CodeSniffer Standards
cakephp/chronos                       1.0.4              A simple API extension for DateTime.
cakephp/debug_kit                     3.5.1              CakePHP Debug Kit
cakephp/migrations                    1.6.7              Database Migration plugin for CakePHP 3.0 based on Phinx
cakephp/plugin-installer              0.0.15             A composer installer for CakePHP 3.0+ plugins.
chialab/questionnaires-converter      dev-master f66eaae Questionnaires converter
composer/semver                       1.4.2              Semver library that offers utilities, version constraint parsing and validation.
dnoegel/php-xdg-base-dir              0.1                implementation of xdg base directory specification for php
doctrine/instantiator                 1.0.5              A small, lightweight utility to instantiate objects in PHP without invoking their constructors
fzaninotto/faker                      v1.6.0             Faker is a PHP library that generates fake data for you.
jakub-onderka/php-console-color       0.1                
jakub-onderka/php-console-highlighter v0.3.2             
jdorn/sql-formatter                   v1.2.17            a PHP SQL highlighting library
mobiledetect/mobiledetectlib          2.8.24             Mobile_Detect is a lightweight PHP class for detecting mobile devices. It uses the User-Agent string combined with specific...
myclabs/deep-copy                     1.5.5              Create deep copies (clones) of your objects
nikic/php-parser                      v3.0.2             A PHP parser written in PHP
[...]

PR #477, opened 2017-01-14
Fix tests in 3.4.0+

markstory:
Apply patch from @fquffio posted in #471 to fix tests and not break compatibility with <3.4.x. This is a much better fix than the hastily pushed bad change I made last night.


antograssiot:
Looks good but tests needs update for 5.5

antograssiot:
I've added a commit that should please PHPUnit but it might not be the best @markstory

markstory:
@antograssiot That works for me.

PR #478, opened 2017-01-15
Add CakePHP version directly in the toolbar

antograssiot:
qisn7gbfew

sorry the gif quality is pretty bad but don't want to spend time on this 😢

### DIFF
--- a/webroot/css/toolbar.css
+++ b/webroot/css/toolbar.css
@@ -68,6 +68,7 @@ p {
 .toolbar.open {
 	border-radius: 0;
 	overflow: auto;
+	padding-right: 100px;
 }
 
 .hidden {
@@ -128,6 +129,13 @@ p {
 
 .history-mode .panel-active .panel-button {
 	color: #ab47bc;
+}
+
+#cake-version {
+	position:absolute;
+	right:40px;
+	cursor: default;
+	padding-right: 0;
 }
 
 /**
--- a/webroot/css/toolbar.css
+++ b/webroot/css/toolbar.css
@@ -131,7 +131,7 @@ p {
 	color: #ab47bc;
 }
 
-#cake-version {
+.cake-version {
 	position:absolute;
 	right:40px;
 	cursor: default;
--- a/src/Template/Requests/view.ctp
+++ b/src/Template/Requests/view.ctp
@@ -28,6 +28,9 @@ use Cake\Core\Configure;
         <?php endif ?>
     </li>
     <?php endforeach; ?>
+    <li id="cake-version" class="panel">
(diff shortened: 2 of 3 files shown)


webroot/css/toolbar.css
```diff
@@ -130,6 +131,13 @@ p {
 	color: #ab47bc;
 }
 
+#cake-version {
```
antograssiot: There are some of those already bot.
Mark let me know if you want me to transform this into a class. no problem

ADmad: @antograssiot You can fix all the ID selectors :)

inoas:
Don't worry, we have to change all those icon/references to that icon to svg at some point anyway.

antograssiot:
@inoas I don't understand your comment.... I don't see the link between the cake version display that I added and icon/svg stuff ?!

inoas:
Ah you meant your animated gif, not the cakephp icon, nvm sorry for the noise.

markstory:
Seems reasonable. I was wondering if there would be value in having a 'packages' panel that listed out the versions of all the packages an application used?

antograssiot:
So the idea would be to parse the composer.lock ?

markstory:
@antograssiot I assume there is something we could use in composer, or by reading the lock file.

antograssiot:
@markstory I've quickly worked on your idea and the result could look like the following.
The tooltip of each row would show the package's description and by clicking on the package name it open the packagist page of the package in a new tab (I found that the homepage info is not reliable but we can also use this or both with minor work if you prefer).

packagist

Is it near what you had in mind? Should I push those changes and do I need to remove the Cakephp version from the toolbar then ?

lorenzo:
@antograssiot I love it! It would be really nice if it could tell you if there is a new version you can install, like showing outdated packages

josegonzalez:
Probably worth putting the packages list in a separate pr so we can get this one merged?

antograssiot:
Except if this one doesn't make sense to others @savant

dereuromark:
People can always disable it locally if they dont want it, right?

antograssiot:
I don't see how @dereuromark but I might have missed something

dereuromark:
I thought it worked like this to disable a tab:
```php
Configure::write(
    'DebugKit.panels',
    [
        foobar => false
    ]
);
```

antograssiot:
Bit it is not a real panel... it is not part of the config

dereuromark:
OK, a new PR is best to not block this one from moving forward while discussing the new feature.

markstory:
@antograssiot That is *AMAZING*

PR #480, opened 2017-01-17
Add a new Packages Panel

antograssiot:
This add the packages Panel as discussed  in #478 

For now it is dead simple, just parsing the `composer.lock` but I love the idea proposed by @lorenzo to get an up-to-date report or something. I will probably try to see how it can be done in the next weeks.

Is this ok to target master ?

b500f8b4-dbe2-11e6-83d9-b6063db7f361

### DIFF
--- a/src/Template/Element/packages_panel.ctp
+++ b/src/Template/Element/packages_panel.ctp
@@ -17,36 +17,61 @@
  * @type array $packages
  */
 ?>
-
-<?php if (!empty($packages)): ?>
-    <?php foreach ($packages as $title => $requirements): ?>
-    <section class="section-tile">
-        <h3><?= __d('debug_kit', $title) ?> </h3>
-        <table cellspacing="0" cellpadding="0" class="debug-table">
-            <thead>
-            <tr>
-                <th>Name</th>
-                <th>Version</th>
-            </tr>
-            </thead>
-            <tbody>
-            <?php foreach ($requirements as $package): ?>
-                <?php extract($package); ?>
-                <tr>
-                    <td title="<?= $description ?>">
-                        <a href="https://packagist.org/packages/<?= $name ?>" title="<?= $description ?>" target="_blank" class="package-link"><?= $name ?></a>
-                    </td>
-                    <td>
-                        <span class="package-version"><span class="panel-summary"><?= $version ?></span></span>
-                    </td>
-                </tr>
-            <?php endforeach; ?>
-            </tbody>
-        </table>
-    </section>
-    <?php endforeach; ?>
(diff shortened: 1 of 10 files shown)


antograssiot:
I don'the like it as is, I will push an update later to remove constants and therefore title from he controller and passed two variable instead of a huge associated array. Nevermore for the loop duplication in the view.

lorenzo:
@antograssiot it is ok to merge in master

src/Template/Element/packages_panel.ctp
```diff
@@ -0,0 +1,52 @@
...
+                <th>Name</th>
+                <th>Version</th>
+            </tr>
+            </thead>
+            <tbody>
+            <?php foreach ($requirements as $package): ?>
+                <?php extract($package); ?>
+                <tr>
+                    <td title="<?= $description ?>">
+                        <a href="https://packagist.org/packages/<?= $name ?>" title="<?= $description ?>" target="_blank" class="package-link"><?= $name ?></a>
```
markstory: Should do HTML escaping on `name` and `description`.

src/Template/Element/packages_panel.ctp
```diff
@@ -0,0 +1,52 @@
...
+                    </td>
+                    <td>
+                        <span class="package-version"><span class="panel-summary"><?= $version ?></span></span>
```
markstory: More HTML escaping here.

src/Template/Element/packages_panel.ctp
```diff
@@ -0,0 +1,52 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+
+/**
+ * @type \DebugKit\View\AjaxView $this
+ * @type array $packages
+ */
+?>
+
+<?php if (!empty($packages)): ?>
```
markstory: Should we roll the current version into this panel, and clean up the toolbar state a bit?

antograssiot: The current version will be here indeed so if you prefer to remove it from the toolbar it works for me.

markstory: I don't think we need both.

antograssiot: Ok I'll do that. Just reverting the other PR should be enough

antograssiot:
ok I've done some cleanup and included the requested changes (HTML escaping and remove the recently added CakePHP version from the toolbar)

capture d ecran 2017-01-18 a 06 09 20

markstory:
Great work @antograssiot I'll get this merged and tagged tonight.

PR #485, opened 2017-01-25
Add CONFIG to CakePHP Constants list

garas:
In Environment panel, CONFIG should be within _CakePHP Constants_ list, not _Application Constants_.

Also remove DS and ROOT from hidden constants list, as they are already in displayed list.

### DIFF
--- a/src/Panel/EnvironmentPanel.php
+++ b/src/Panel/EnvironmentPanel.php
@@ -59,14 +59,14 @@ class EnvironmentPanel extends DebugPanel
             'WWW_ROOT' => WWW_ROOT
         ];
 
-        $cakeConstants = array_fill_keys(
+        $hiddenCakeConstants = array_fill_keys(
             [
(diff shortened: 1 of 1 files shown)


src/Panel/EnvironmentPanel.php
```diff
@@ -48,6 +48,7 @@ protected function _prepare()
             'CACHE' => CACHE,
             'CAKE' => CAKE,
             'CAKE_CORE_INCLUDE_PATH' => CAKE_CORE_INCLUDE_PATH,
+            'CONFIG' => CONFIG,
```
markstory: This constant is defined in the application, which is why it was an application constant.

garas: But then only DS, TIME_START and time constants are _define()_'ed withing cakephp/cakephp (except `tests/bootstrap.php`).

Every other constant is defined in `cakephp/app` bootstrap.php.

CakePHP depends on CONFIG in multiple places, so it should be core constant.

It listed as core constant in [docs](https://book.cakephp.org/3.0/en/core-libraries/global-constants-and-functions.html#core-definition-constants)

markstory: Fair enough.

PR #486, opened 2017-01-25
Add space between array key and value when copying 'neat array' to clipboard

garas:
`makeNeatArray` is used in Variables, Request, Session, Includes panels.

If you try to copy browser result to clipboard, text is hard to read.

When copying list to clipboard, result is
```
keyvalue1
keyvalue2
```

After the fix, clipboard contents are
```
key value1
key value2
```


markstory:
Thanks!

PR #487, opened 2017-01-27
Add a tool to check composer outdated dependencies

antograssiot:
You can check fot all outdated dependencies or choose to select only direct dependencies.

This is not ready to be merged, I'll try to find time to add tests but wanted to get feedback already.
There are places where I should add translation also if possible.

The gif screen capture is way slower than the actual behavior as it's recorded online.
zsfkhkzhzn

### DIFF
--- a/src/Template/Element/packages_panel.ctp
+++ b/src/Template/Element/packages_panel.ctp
@@ -22,6 +22,11 @@
         <?= __d('debug_kit', '{0} not found', "'composer.lock'"); ?>
     </div>
 <?php else: ?>
+    <div class="check-update">
+        <button class="btn-primary">Check for Updates</button>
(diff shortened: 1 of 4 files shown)


lorenzo:
This looks great!

src/Template/Element/packages_panel.ctp
```diff
@@ -75,3 +80,60 @@
...
+        }
+
+        function buildSuccessfulMessage(response) {
+            var html = '';
+            if (response.packages.bcBreaks === undefined && response.packages.semverCompatible === undefined) {
+                return '<pre class="success-message">All dependencies are up to date</pre>';
+            }
+            if (response.packages.bcBreaks !== undefined) {
+                html += '<h4>Update with potential BC break</h4>';
+                html += '<pre>' + response.packages.bcBreaks + '</pre>';
```
antograssiot: I'm actually using the tags issued by the console output to style the result so escaping here would break that.

Issue #489, opened 2017-02-03
Add configuration to disable/reorder panels


nrother:
Currently all default panels are always shown. With the recent addition of the new panels the list got so long that it overflows to the next line on my small monitor I usually keep firefox on.

As far as I can see one can add custom panels using the `DebugKit.panels` configuration, but there is no way to remove the default panels.

dereuromark:
I always thought

    'MyPanel' => false

would disable that one.
Maybe some docs missing if that works?
Otherwise I agree, this would be useful. 

Reordering might be a bit more tricky.

lorenzo:
We just need to make the bar scroll horizontally on overflow

nrother:
Actually I also expected that to work, but it did not... [The code](https://github.com/cakephp/debug_kit/blob/master/src/ToolbarService.php#L134) also does not look like if that would be supported.

Horizontal scrolling isn't much better that vertical scrolling, both keep the variables panel, I use most often, out of reach :frowning:  (Actually I think that horizontal scrolling is worse...)

dereuromark:
@nrother You can make a PR to allow the disabling etc, should be a quick addition.

lorenzo:
@nrother we can also make the variables panel the first one as it is the most commonly used

nrother:
@dereuromark Actually I'm not quite sure how to do this in a elegant way... A simple 
```php
foreach ($this->config('panels') as $panel => $enabled) {
	if ($enabled) {
	    $this->registry->load($panel);
    }
}
```
would not work if someone tries to add a panel while disabling another one (like `'panels' => ['Custom', 'DebugKit.Timers' => false]` because the panel name is now sometimes the key and sometimes the value... Is there an elegant way to solve this?

dereuromark:
Yes, other places in core and plugins check on the $panel being numeric here.
If so, then they are only values no key.

markstory:
Closing as a pull request is open now.

PR #490, opened 2017-02-06
Add option to disable build-in panels

nrother:
Ref #489


dereuromark:
This will need a test.

tests/TestCase/ToolbarServiceTest.php
```diff
@@ -73,6 +73,25 @@ public function testLoadPanels()
...
+    /**
+     * Test disabling panels.
+     *
+     * @return void
+     */
+    public function testDisablePanels()
+    {
+        $bar = new ToolbarService($this->events, ['panels' => [
+            'DebugKit.SqlLog' => false,
+            'DebugKit.SqlLog' => true,
```
dereuromark: Twice the same key? typo probably.

nrother:
Ok, this is not working as it should:
```
>>> $bar = new ToolbarService(new \Cake\Event\EventManager(), ['panels' => ['DebugKit.Log' => false]]);
=> DebugKit\ToolbarService {#228}
>>> $bar->config()                                                              => [
     "panels" => [
       0 => "DebugKit.Cache",
       1 => "DebugKit.Session",
       2 => "DebugKit.Request",
       3 => "DebugKit.SqlLog",
       4 => "DebugKit.Timer",
       5 => "DebugKit.Log",
       6 => "DebugKit.Variables",
       7 => "DebugKit.Environment",
       8 => "DebugKit.Include",
       9 => "DebugKit.History",
       10 => "DebugKit.Routes",
       11 => "DebugKit.Packages",
       12 => "DebugKit.Mail",
       "DebugKit.Log" => false,
     ],
     "forceEnable" => false,
   ]
```
Passing `['DebugKit.Log' => false]` does not overwrite the old value but creates a new key. Is this kind of problem also solved in Cake somewhere?

dereuromark:
There are different solutions, one for example merging in the defaults via "Key"= >true pair etc.
That should solve the issue afaik.

tests/TestCase/ToolbarServiceTest.php
```diff
@@ -73,6 +73,25 @@ public function testLoadPanels()
...
+     * Test disabling panels.
+     *
+     * @return void
+     */
+    public function testDisablePanels()
+    {
+        $bar = new ToolbarService($this->events, ['panels' => [
+            'DebugKit.SqlLog' => false,
+            'DebugKit.Cache' => true,
+            'DebugKit.Session'
```
markstory: Will this setup allow people to only disable the panels they want? What happens if a panel has both a numeric index and the same panel name is set to `false` as a key?

nrother: All panels are enabled by default (due to the `$_defaultConfig`). So you explicitly need to disable panels you don't want. Only custom panels need to be included/enabled (as before).

When a panel is contained with a numeric key **and** set to `false` it will be loaded, but this can only happen if the user includes something like `'panels' => ['Test', 'Test' => false]` in their `config/app.php` which *should* not happen...

dereuromark: Looks BC to me. Nice work.

tests/TestCase/ToolbarServiceTest.php
```diff
@@ -73,6 +73,25 @@ public function testLoadPanels()
         $this->assertGreaterThan(1, $this->events->listeners('Controller.shutdown'));
         $this->assertInstanceOf('DebugKit\Panel\SqlLogPanel', $bar->panel('SqlLog'));
     }
+
```
dereuromark: Can you fix this please? Make sure your IDE has trim right enabled to not pollute the files with whitespace noise.

dereuromark:
👍 Let's ship it then :)

PR #491, opened 2017-02-06
Fix phpunit issue

dereuromark:
Dont use the broken global one

Fixes at least the 5.6 fails.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     },
     "require-dev": {
         "cakephp/cakephp-codesniffer": "~2.1",
-        "phpunit/phpunit": "5.*"
+        "phpunit/phpunit": "<6.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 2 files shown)


ravage84:
Wouldn't ``composer  test`` work, if we had a ``test``script defined under the ``script`` key in ``composer.json``?

dereuromark:
It might, but thats not for travis IMO, thats for end user "usage" :)

composer.json
```diff
@@ -30,7 +30,7 @@
     },
     "require-dev": {
         "cakephp/cakephp-codesniffer": "~2.1",
-        "phpunit/phpunit": "5.*"
+        "phpunit/phpunit": "<6.0"
```
ravage84: Would the tests work with PHPUnit 3.7? 😼
I think not. So something like ``>=4.0 <6.0`` would be _more_ correct.

antograssiot:
All tests passing now

composer.json
antograssiot: What's the point ? I don't understand as composer will always fetch the latest version possible and debug_kit is php 5.5...

PR #497, opened 2017-02-15
minor fixes to unit tests

gemal:


Test/Case/Lib/DebugMemoryTest.php
```diff
@@ -46,18 +46,18 @@ public function testMemorySettingAndGetting() {
 		$this->assertTrue($result);
 
 		$result = DebugMemory::getAll(true);
-		$this->assertEquals(count($result), 1);
+		$this->assertCount(1, $result);
```
gemal: I didn't change tabs or spaces here

AD7six: The message isn't saying you added tabs - it's pointing out that the modified lines (or simply, the file) are tab indented. That is something which needs addressing.

PR #500, opened 2017-02-17
Replace deprecated App::pluginPath with CakePlugin::path

tersmitten:


Test/Case/Controller/Component/ToolbarComponentTest.php
```diff
@@ -160,7 +160,7 @@ public function testLoadPanelsError() {
  * @return void
  */
 	public function testLoadPluginPanels() {
-		$debugKitPath = App::pluginPath('DebugKit');
+		$debugKitPath = CakePlugin::path('DebugKit');
```
dereuromark: Stickler seems to not understand 2.x here.

tersmitten: I also thought so, bad stickler-ci...

markstory:
I'll get the travis and stickler issues fixed up.

PR #502, opened 2017-02-25
Update travis tests.

markstory:
* Test against more recent versions of CakePHP.
* Test against PHP 5.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,13 @@ php:
   - 5.3
   - 5.4
   - 5.5
+  - 5.6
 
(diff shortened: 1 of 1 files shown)


ravage84:
@markstory according to the [requirements](https://github.com/cakephp/debug_kit/tree/2.2.6#requirements), the minimum CakePHP version is 2.2.x. Is this still true? If so, is this still covered by Travis? If it is not covered, shouldn't we bump the minimum requirement or at least issue a notice that it is not officially supported anymore?

PR #505, opened 2017-03-04
Pass plugin to the iframe for rendering mail messages.

### LINKED ISSUE #501: Mailer preview, preview iframe missing plugin query param when using MailerPreviews from plugin
When using a MailerPreview from a plugin the iframe loading the preview does not get the plugin query param, and fails to find the correct MailerPreview class

sitedyno:
Otherwise mail previews from plugins would not be found.

Fixes #501


markstory:
Thanks!

PR #507, opened 2017-03-06
Ignore log files genearated by tests

sitedyno:
Files are created in the root directory of the project when running tests outside of an app. So easy to accidentally add them to the project.


markstory:
Thanks!

Issue #512, opened 2017-03-20
Data too long for column 'content' at row 1


nhristov:
`SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'content' at row 1`

My fix was simple:
```sql
ALTER TABLE `panels` CHANGE `content` `content` LONGBLOB NULL DEFAULT NULL;
```

markstory:
Pull request up now.

PR #514, opened 2017-03-25
Make content column a longblob.

markstory:
Try to avoid truncation issues for more folks.

Refs #512

### DIFF
--- a/tests/Fixture/PanelsFixture.php
+++ b/tests/Fixture/PanelsFixture.php
@@ -12,7 +12,7 @@
  */
 namespace Debugkit\Test\Fixture;
 
-use Cake\Database\Schema\TableSchema;
+use Cake\Database\Schema\Table as TableSchema;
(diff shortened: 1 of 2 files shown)


tests/Fixture/PanelsFixture.php
```diff
@@ -42,7 +43,7 @@ class PanelsFixture extends TestFixture
         'title' => ['type' => 'string'],
         'element' => ['type' => 'string'],
         'summary' => ['type' => 'string'],
-        'content' => ['type' => 'binary'],
+        'content' => ['type' => 'binary', 'length' => TableSchema::LENGTH_LONG],
```
markstory: Thanks for finding in which version it was added. I will update the minimum version constraints.

antograssiot:
Now it's using chronos 0.1.0. 
The easier would be to require a chronos version  in debug_kit.
Latest version of cakephp/cakephp doesn't réduire `*` anymore.

PR #517, opened 2017-03-31
Add feature to scroll panels

rchavik:
If you have many panels, some panels may be positioned below the visible area.
This change adds scrolling capability to allow viewing these hidden panels.

### DIFF
--- a/src/Template/Requests/view.ctp
+++ b/src/Template/Requests/view.ctp
@@ -16,6 +16,17 @@ use Cake\Core\Configure;
 </div>
 
 <ul id="toolbar" class="toolbar">
+     <li id="panel-button-left" class="panel hidden">
+        <span class="panel-button">
+            &lsaquo;
+        </span>
+    </li>
+    <li id="panel-button-right" class="panel hidden">
+        <span class="panel-button">
+            &rsaquo;
+        </span>
+    </li>
+    <div class="toolbar-inner">
     <?php foreach ($toolbar->panels as $panel): ?>
     <li class="panel hidden" data-id="<?= $panel->id ?>">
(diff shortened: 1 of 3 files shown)


src/Template/Requests/view.ctp
```diff
@@ -16,6 +16,17 @@ use Cake\Core\Configure;
...
 
 <ul id="toolbar" class="toolbar">
+     <li id="panel-button-left" class="panel hidden">
+        <span class="panel-button">
+            &lt;
+        </span>
+    </li>
+    <li id="panel-button-right" class="panel hidden">
+        <span class="panel-button">
+            &gt;
```
markstory: What about the [fancy unicode left and right chevrons](http://www.fileformat.info/info/unicode/char/203a/index.htm)?

rchavik: Changed :)

rchavik:
Anything else to get this merged?

markstory:
@rchavik Just going to test it out locally.

rchavik:
Thanks.

I was confused when I couldn't find the Timers panel :)

markstory:
I have the same problem on my smallish laptop screen. Thanks for fixing this before I could :clap:

webroot/css/toolbar.css
```diff
@@ -110,10 +117,14 @@ p {
 	vertical-align: middle;
 }
 
+#panel-button-right {
```
ravage84: @markstory it would help,  if stickler-ci would provide a resource link, which explains why.

markstory: I can see if the underlying linter can do that.

PR #519, opened 2017-04-25
Replace death star version constraint

ravage84:
https://thephp.cc/news/2017/02/death-star-version-constraint


markstory:
Thanks!

Issue #494, opened 2017-02-08
Build url in template Previewing Mailer class will result in DebugKit plugin routing


tenkoma:
This is a bug in Mail Preview.

* Version CakePHP 3.3.14, DebugKit 3.8.0

### What you did

The code is as follows

```php
<?php
// src/Mailer/UserMailer.php
namespace App\Mailer;

use Cake\Mailer\Mailer;

class UserMailer extends Mailer
{
    public function welcome()
    {
        return $this->to("yourname@example.com")
            ->subject("welcome!")
            ->template("welcome_mail");
    }
}

```

```php
<?php // src/Template/Email/text/welcome_mail.ctp ?>
<?= $this->Url->build(
    ['controller' => 'Pages', 'action' => 'index'],
    ['fullBase' => true]
) ?>
```

```php
<?php
// src/Mailer/Preview/UserMailPreview.php
namespace App\Mailer\Preview;

use DebugKit\Mailer\MailPreview;

class UserMailPreview extends MailPreview
{
    public function welcome()
    {
        return $this->getMailer("User")->welcome();
    }
}

```

### What happened

Occurred `Cake\Routing\Exception\MissingRouteException`. Error Message:

```
Error: A route matching "array ( 'controller' => 'Pages', 'action' => 'index', 'plugin' => 'DebugKit', '_ext' => NULL, )" could not be found.
```

### What you expected to happen

Render url likes `http://example.com/pages`

### Temporary workaround

Include `'plugin' => false` in routing array.

markstory:
Looks like this was fixed back in #614.

PR #523, opened 2017-05-06
WIP: Fix build url error in mail preview text

tenkoma:
Fix: #494

### DIFF
--- a/src/Mailer/PreviewResult.php
+++ b/src/Mailer/PreviewResult.php
@@ -13,6 +13,7 @@
 namespace DebugKit\Mailer;
 
 use Cake\Mailer\Mailer;
+use Cake\Routing\Router;
 use ReflectionClass;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Mailer/PreviewResult.php
```diff
@@ -55,7 +56,9 @@ protected function processMailer(Mailer $mailer, $method)
             ->getMethod('_renderTemplates')
             ->getClosure($email);
 
+        Router::setRequestInfo(['plugin' => null]);
```
chinpei215: Sorry if I am misunderstanding, but I don't think this patch fixes the issue. Even after your fix, it wouldn't work correctly if you want to preview emails from plugins or prefixed actions,. For example, if you send an email in an admin action, a URL array `['controller' => 'Pages', 'action' => 'index']` must be `/admin/pages`, but it will be `/pages` in the mail preview. It is not correct. We would have to add a feature that allows people to tell the original request context to DebugKit, instead of this kind of hack.

tenkoma: @chinpei215 Thanks to review. Your point is collect.

> We would have to add a feature that allows people to tell the original request context to DebugKit, instead of this kind of hack.

I see.

tenkoma:
I will read the code more, thanks.

PR #524, opened 2017-05-07
Add Middleware & fix a few bugs

markstory:
Add a middleware version of debug kit, and fix a few problems I found while testing these changes out.

#### Fixed bugs

* Scroll buttons would hide expanded panels.
* History view was incorrect as panel content was shifted 2 panels to the left because of the new scroll buttons.


composer.json
```diff
@@ -23,7 +23,7 @@
         "source": "https://github.com/cakephp/debug_kit"
     },
     "require": {
-        "cakephp/cakephp": "^3.2.7",
+        "cakephp/cakephp": ">=3.3.0,<4.0.0",
```
dakota: I think `^3.3.0` would be the same?

markstory: Yes it would. Thanks!

Issue #528, opened 2017-05-11
DebugKit breaks plugin static serving


highstrike:
~~The following commit https://github.com/cakephp/debug_kit/commit/aec9876daa0515c1776db4fc59c260ead1a88d3e~~ broke the ability to serve static content from a plugin with query parameters.

(not sure if this is actually the commit that broke it because I haven't run `composer update` in a few weeks, so I might have been on 3.9.3, with cakephp 3.4.5) I am now on 3.10.1 with cakephp 3.4.6

For example I have a plugin called 'hello/kitty'. I then load said plugin with `Plugin::load('Hello/Kitty');` in my bootstrap file.

My plugin then contains a file called `index.html` inside the webroot/img directory.
Normally the path to it would be `http://domain.tld/hello/kitty/img/index.html`.

This works fine.

However, as soon as I change the request to `http://domain.tld/hello/kitty/img/index.html?page=1` I get "Stream is not writable - RuntimeException"

I don't even have to put the whole page query either, just `?` is enough to break it
`http://domain.tld/hello/kitty/img/index.html?`

The query param is actually used by javascript, I have a script there that does something 😄

Please see attached screenshot.

![image](https://cloud.githubusercontent.com/assets/2379538/25965220/ff29dc00-368f-11e7-9f62-f219ecc607fd.png)

markstory:
Thanks! I'll see if I can reproduce this issue.

markstory:
Pull request open now.

PR #529, opened 2017-05-12
Fix errors on non-writable streams.

markstory:
Don't try to modify read only streams like file responses.

Refs #528


highstrike:
Sweet, you fixed it, thanks!

PR #542, opened 2017-08-03
Fix empty email previews (#541)

### LINKED ISSUE #541: Email preview doesn't work with default template
I can't get the email preview to work unless I set the template in the Preview class or in the Mailer.

I've found out the problem to be in line https://github.com/cakephp/debug_kit/blob/master/src/Mailer/PreviewResult.php#L45

This line tries to check that the Mailer doesn't already have a template set and sets the template to the method name. This fails because the ```Mailer::template()``` call will always return the Mailer itself since it's implemented as a magic method that will call the method with the same name in the email class then return always the Mailer for method chaining.

berarma:
This fixes #541.

### DIFF
--- a/src/Mailer/PreviewResult.php
+++ b/src/Mailer/PreviewResult.php
@@ -42,14 +42,14 @@ class PreviewResult extends AbstractResult
      */
     protected function processMailer(Mailer $mailer, $method)
     {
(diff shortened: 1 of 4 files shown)


markstory:
Thanks!

Issue #551, opened 2017-09-29
Hide DebugKits Routes


dereuromark:
In the Routes tab the debug kit routes are right the first ones.
Those are usually the ones least interested in usually :)
Probably best to hide them, or at least put them to the end and display in greyish.

markstory:
Closing as a pull request is open.

PR #552, opened 2017-09-29
Use 3.1+ dashed route scheme for URLs.

dereuromark:
Resolves https://github.com/cakephp/debug_kit/issues/551

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -2,10 +2,10 @@
 use Cake\Routing\RouteBuilder;
 use Cake\Routing\Router;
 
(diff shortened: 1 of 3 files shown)


dereuromark:
Sth is not yet quite working out regarding JS+CSS on the toolbar at the bottom of the page.
Must be missing one tweak. 

This is probably better to go into a minor instead of patch release. But as require-dev shouldnt be too much of an issue.

markstory:
I'll get this and the other pull request merged in and fix any gaps afterwards.

PR #553, opened 2017-09-29
Hide debug-kit noise routes.

dereuromark:
Resolves https://github.com/cakephp/debug_kit/issues/551

### DIFF
--- a/src/Template/Element/routes_panel.ctp
+++ b/src/Template/Element/routes_panel.ctp
@@ -15,6 +15,9 @@ $routes = Cake\Routing\Router::routes();
     </thead>
     <tbody>
     <?php foreach ($routes as $route): ?>
+        <?php if (strpos($route->getName(), 'debugkit.') === 0) {
+            continue;
+        }?>
         <tr class="<?= ($matchedRoute === $route->template) ? 'highlighted' : '' ?>">
             <td><?= h(\Cake\Utility\Hash::get($route->options, '_name', $route->getName())) ?></td>
             <td class="left"><?= h($route->template) ?></td>


src/Template/Element/routes_panel.ctp
```diff
@@ -15,6 +15,9 @@ $routes = Cake\Routing\Router::routes();
     </thead>
     <tbody>
     <?php foreach ($routes as $route): ?>
+        <?php if (strpos($route->getName(), 'debugkit.') === 0) {
+            continue;
```
markstory: Perhaps we should output them in a muted color instead of hiding entirely?

lorenzo: That sounds better

dereuromark: then u need to scroll one page. why would anyone be interested in the help tools routes? maybe have a toggle  button to include them and by default exlude? or move to the end.

ravage84: How about a configuration setting that is enabled or disabled by default which hides them as @dereuromark proposes?

markstory: Scrolling is dependent on the number of routes in the app. I don't think we can make decisions based on whether or not people will need to scroll.

Moving debugkit routes to the end is an option as well.

dereuromark: So are we OK with at least graying them out? #999 or sth?

markstory: Displaying in grey sounds good to me.

ADmad:
@dereuromark Ping

dereuromark:
I am fine if someone wants to finish it for us.

saeideng:
an alternate PR merged by mars story

PR #557, opened 2017-10-16
Use  DIRECTORY_SEPARATOR for file path

stmeyer:
At the moment you get on windows something like `CAKE/Datasource\ConnectionManager.php`


dereuromark:
thank you.

PR #560, opened 2017-11-08
Don't try to load routes if router not already initialized.

### LINKED ISSUE #559: Unable to get rid of routes.php & setup routes through Application::routes() only.
I was trying out loading routes through `Application::routes()` without having `config/routes.php` and things work as expected until I loaded `DebugKit`.

The problem occurs because `DebugKitMiddleware` get's added to the queue as first middleware and when it's loading it `RoutingPanel` the routes have still not been loaded as `RoutingMiddleware` which calls `Application::routes()` has still not run. Hence the `Router` tries to load `config/routes.php` resulting in error.

Since we indent to do away with `config/routes.php` in future this issue needs to be fixed.

ADmad:
Fixes #559.

### DIFF
--- a/src/Panel/RoutesPanel.php
+++ b/src/Panel/RoutesPanel.php
@@ -13,6 +13,7 @@
  */
 namespace DebugKit\Panel;
 
+use Cake\Core\Configure;
 use Cake\Event\Event;
 use Cake\Routing\Router;
 use DebugKit\DebugPanel;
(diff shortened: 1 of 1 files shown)


src/Panel/RoutesPanel.php
```diff
@@ -30,6 +31,11 @@ class RoutesPanel extends DebugPanel
      */
     public function summary()
     {
+        $appClass = Configure::read('App.namespace') . '\Application';
+        if (class_exists($appClass, false) && !Router::$initialized) {
```
saeideng: https://github.com/cakephp/cakephp/blob/80e225590d06e7e40ea6aab21571269c7873c3a5/src/Routing/Router.php#L41-L42

markstory: @saeideng We'll have $initialized around until 4.0.0, as we can't remove the autoloading of routes until the.

ADmad:
@markstory ping

PR #563, opened 2017-11-28
Let variables panel test serialization and handle debug info exceptions.

ndm2:
I stumbled over a weird problem, that is, a JSON request in which an exception occurrs would somehow respond with an HTML error page instead of serialized JSON data, but only for some specific type of exception, in this case for an `InvalidArgumentException` caused by a non-existing association ("x is not associated with y").

The problem was that the variables panel invokes `__debugInfo()` on the the erroneous query, where the same exception would be triggered again as a result of `Query::sql()` being invoked, which then in turn caused the already rendered but not yet sent JSON response to be "overwritten" with the same error but this time in HTML format.

This patch furthermore applies some serialization testing that is currently also done on all data in `ToolbarService::saveData()`. This allows to present at least a partial variable tree, instead of swallowing everything. This may not be an ideal solution, and maybe it would be better to implement something similar for use in `ToolbarService::saveData()` instead, not quite sure about that?

### DIFF
--- a/src/Panel/VariablesPanel.php
+++ b/src/Panel/VariablesPanel.php
@@ -62,6 +62,32 @@ class VariablesPanel extends DebugPanel
     }
 
     /**
+     * Safely retrieves debug information from an object
+     * and applies a callback.
+     *
+     * @param callable $walker The walker to apply on the debug info array.
+     * @param object $item The item whose debug info to retrieve.
+     *
+     * @return array|string
+     */
+    protected function _walkDebugInfo(callable $walker, $item)
+    {
+        try {
+            $info = $item->__debugInfo();
+        } catch (\Exception $exception) {
+            return sprintf(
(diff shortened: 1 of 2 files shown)


src/Panel/VariablesPanel.php
```diff
@@ -103,7 +129,23 @@ public function shutdown(Event $event)
                 );
             } elseif (is_object($item) && method_exists($item, '__debugInfo')) {
                 // Convert objects into using __debugInfo.
-                $item = array_map($walker, $item->__debugInfo());
+                $item = $this->_walkDebugInfo($walker, $item);
+            }
+
+            if (is_object($item) ||
+                is_resource($item)
```
markstory: Resources aren't generally serializable. Perhaps we should just print the resource type instead?

ndm2: Good point, I mixed things up, showing that it's a resource alongside with its type sounds like a good idea.

PR #565, opened 2017-12-11
Test against PHP 7.2

carusogabriel:


markstory:
Looks like a few tests are failing.

carusogabriel:
Yes, because [now `count` warnings if the parameters isn't an `array` or `Object` that implements `Countable`](http://php.net/manual/en/function.count.php#refsect1-function.count-changelog). Gonna work on that.

josegonzalez:
Weird, `ArrayObject` [implements Countable](https://secure.php.net/manual/en/class.arrayobject.php)...

carusogabriel:
@josegonzalez I'm gonna use [ArrayObject's `count` attribute](http://php.net/manual/en/arrayobject.count.php) instead.

markstory:
Thanks!

PR #571, opened 2017-12-21
adds a timestamp to toolbar.js and updates test

### LINKED ISSUE #570: ToolbarService::injectScripts should timestamp JS file to fix caching conflicts
I was running a previous version of DebugKit and updated to the latest. After doing this the DebugKit icon on the app's pages failed to load. Inspecting the `<iframe>` source code revealed a routing not found error which sent me down the wrong path to fix it.

My browser had cached an outdated version of `toolbar.js`. Once I cleared the cache the DebugKit bar appeared.

The `ToolbarService::injectScripts` should append a timestamp taken from the `toolbar.js` file. This should be added as a query parameter `/debug_kit/js/toolbar.js?384773372` (just a unix timestamp).

This will resolving cache issues whenever the file is changed.

thinkingmedia:
Closes #570 

This PR adds a timestamp to the `toolbar.js` file which resolves caching issues in the browser.

I've tried to handle the edge case were the developer has copied the files locally to their webroot. What I don't do is verify if the two files are the same. I think that's a completely different issue.

### DIFF
--- a/src/ToolbarService.php
+++ b/src/ToolbarService.php
@@ -218,12 +218,13 @@ class ToolbarService
             str_replace('/', DIRECTORY_SEPARATOR, WWW_ROOT . 'debug_kit/' . $url),
             str_replace('/', DIRECTORY_SEPARATOR, Plugin::path('DebugKit') . 'webroot/' . $url)
         ];
-        $url = '/debug_kit/'.$url;
-        foreach($filePaths as $filePath) {
+        $url = '/debug_kit/' . $url;
+        foreach ($filePaths as $filePath) {
             if (file_exists($filePath)) {
                 return $url . '?' . filemtime($filePath);
             }
         }
+
         return $url;
     }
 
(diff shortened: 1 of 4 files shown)


thinkingmedia:
Oh stickler I forgot about you....

markstory:
Few more tests are failing, but otherwise this looks good.

thinkingmedia:
@markstory I forgot to configure phpunit before running the tests. So they all passed. That itself should be a bug.

thinkingmedia:
Why is codecov showing red failures?

markstory:
> Why is codecov showing red failures?

It does that sometimes.

> I forgot to configure phpunit before running the tests. So they all passed.

What did you have to configure? The travis builds don't do a ton of setup beyond what is in the repo.

thinkingmedia:
@markstory if you don't make a copy of `phpunit.xml.dist` to `phpunit.xml` and enable a DB connection there are tests that will be skipped. These are shown on the phpunit output as `S`. I think that's fine. I'm just **rusty** and haven't done PRs in a while.

PR #572, opened 2017-12-21
Add support for outputting formatted SQL from a Query object (issue 568)

### LINKED ISSUE #568: Ability to see SQL generated statement with parameters substituted
_From @brettl on June 11, 2017 19:47_

* [x] enhancement

* CakePHP Version: 3.next
When running the ->sql() function on an ORM call allow the ability to see the sql statement with the parameters in the place holders.

Currently when calling sql on the ORM you will see example "where id = :c1"
allow the ability to call an example ->sql(1) or ->sql()->substituted(); and the resulting sql will now be "where id = 1"
this would allow the ability to take your sql being generated and run in a sql viewer /editor and run your command and possibly see if you are producing the incorrect data or if you are generating the incorrectsql statement

_Copied from original issue: cakephp/cakephp#10761_

thinkingmedia:
Closes #568 

This PR adds support for outputting formatted SQL from a Query object.

Adds the `DebugSql` class that handles the formatting, and unit tests for that class.

The `bootstrap.php` was updated to expose two new functions `sql()` and `sqld()` which work like `pr()` and `dd()`.

Basic usage:

```
$query = $article->find()->where([....]);
sql($query);
```

The above will dump the SQL as formatted HTML and you can call `sqld()` if you PHP to die afterwards.

This is a very handy feature.


thinkingmedia:
Stickler passes green but still shows line ends as errors. I don't see those missing in my editor.

src/DebugSql.php
```diff
@@ -0,0 +1,188 @@
...
+        $sql = (string)$query;
+        if ($showValues) {
+            $sql = method_exists($query, 'getValueBinder')
+                ? static::interpolate($sql, $query->getValueBinder()->bindings())
+                : static::interpolate($sql, $query->valueBinder()->bindings());
+        }
+
+        $trace = Debugger::trace(['start' => 1, 'depth' => $stackDepth + 2, 'format' => 'array']);
+        $file = $trace[$stackDepth]['file'];
+        $line = $trace[$stackDepth]['line'];
```
markstory: You should ensure that the depth key is defined before using it.

thinkingmedia: @markstory  You mean something like `$stackDepth = is_int($stackDepth) ? $stackDepth : 0`?

markstory: No more like an `isse($trace[$stackDepth])`. This method is public and someone could send you `-10` as the depth.

dereuromark: I think he means more the value behind the key: `$trace[$stackDepth]` could be undefined, right?

thinkingmedia: oh right, that makes sense. I'll fix it.

thinkingmedia:
If this gets merged, then I can move code from my `SqlTraceTrait` to the `DebugSql` and that will help make the trait testable. See https://github.com/cakephp/debug_kit/pull/574

src/DebugSql.php
```diff
@@ -77,8 +77,8 @@ public static function sql(Query $query, $showValues = true, $showHtml = null, $
         }
 
         $trace = Debugger::trace(['start' => 1, 'depth' => $stackDepth + 2, 'format' => 'array']);
-        $file = $trace[$stackDepth]['file'];
-        $line = $trace[$stackDepth]['line'];
+        $file = isset($trace[$stackDepth]) ? $trace[$stackDepth]['file'] : 'n/a';
```
markstory: You might want to check that there is a file/line key set. Not all points in a stack trace will have file/line sadly.

thinkingmedia: done.

thinkingmedia:
@markstory stickler is not removing resolving the issues related to linefeeds at the end of the file. Even when stickler-ci gives a pass "no lint errors found". The other stickler issues were marked resolved. Seems to maybe be a bug.

thinkingmedia:
I'm done.

I'm not sure if you guys are waiting for me to do something?

PR #574, opened 2017-12-22
adds a trait that injects a file name comment into SQL queries

### LINKED ISSUE #573: Show the originating PHP file in the SQL log
## Overview

When we execute a `pr("some data here")` to display debug messages the output includes the file path to the PHP file that generated the output.

When we look at the SQL log panel in the DebugKit we don't get the context of the query. We just know it was executed, how many results and the duration. As a programmer, I ask myself this "Why is the SQL being executed?" and I don't know the answer. You either recognize the query or you don't.

The originating PHP file would help, but it's not that simple.

## Challenges

You can't easily tell which PHP file in a stack trace is the true author responsible for creating the query. Furthermore, a programmer might argue he/she prefers to see the Controller instead of the Model, but both are in the call stack.

So we should include a stack trace, but stack traces are really big. We could *reduced* the stack trace (filtering out CakePHP core files) and attach it to the Query as metadata. This would be a configured feature (i.e. debug only) as it has obvious performance impacts. Where and when we do this is open to debate. If we do this in the Query constructor (obvious location) then we're adding core changes to one of the most used objects. So I prefer to generate this data and attach it to the Query object, and if somehow make this purely a DebugKit activity.

The problem here is that we want to track Query creation and not execution. There are no events or behaviors in CakePHP that will allow you to monitor those moments.

## CakePHP Core

I don't see how something like this could be implemented unless changes were made in the CakePHP core. Specially, how do we attach metadata to a Query object? How do we monitor the creation of Query objects?

Some possible approaches:

- event model: Change the CakePHP core to broadcast events about Query object creation.
- factory: Change CakePHP core to use a Query factory class, and configure a DebugKit factory class.
- extension: Add a Table class to DebugKit that logs Query creation, and tell people to `extend` this Table class.
- mixin: Same as above, but as a mixin.
- behavior: Add a DebugKit behavior that *mutates* the owning Table instance and replaces the `find()` method with it's own (a kind of mocking behavior).
- unknown: Something I've missed that maybe uses one of the above (i.e. an event listener that attaches behaviors to Tables automatically, and hacks what we need)

## Feedback

I'm looking to experiment with some of the above ideas, and hope to hear from other people different approaches or opinions about such changes.

thinkingmedia:
closes #573 

Adds a new trait called `SqlTraceTrait` that injects a SQL comment containing the source file that created a new Query object. When you apply the trait to a Table class it overrides the `function query()` method and calls `$query->epilog($comment)` to inject the name of the file that created the query.

It works by finding the first file in a call stack that belongs to the application, and defaults to the last core source file if none found (i.e. I saw this happen for Associations).

sql-log-example

The comments appear in the SQL log as seen in the above image. Since this becomes part of the SQL statement the comments also get captured in the SQL server's log (possible feature side effect). The SQL log panel already applies syntax highlight so the comment has a nice gray color and doesn't visually interfere with the reading of the logs.

The comment is attached to the query using `epilog()` so it will be overwritten if the developer uses `epilog()` for something else. The goal here is to help people see where queries are created, and if you're using `epilog()` for something else, then you're likely to remember why.

The only side side is that you have to manually add this trait to your tables. I happen to have a shared base class named `AppTable` that would add this to all tables, but I know most people don't do this. So I propose if this gets merged into DebugKit that we update the bake project and application template to include this trait. Which will help people see this as a new feature.

If anyone has a better approach, please let me know.

### DIFF
--- a/src/Model/Table/SqlTraceTrait.php
+++ b/src/Model/Table/SqlTraceTrait.php
@@ -8,6 +8,7 @@
  *
  * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
  * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.11.5
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 namespace DebugKit\Model\Table;
--- a/src/Model/Table/SqlTraceTrait.php
+++ b/src/Model/Table/SqlTraceTrait.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace DebugKit\Model\Table;
+
+use Cake\Core\Configure;
+use Cake\Error\Debugger;
+use Cake\ORM\Query;
+
+/**
+ * This trait modifies a Table class to inject metadata into a Query object that can be used to provide a call track
+ * in the SQL log panel.
+ *
+ * @mixin \Cake\ORM\Table
+ */
+trait SqlTraceTrait
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function query()
+    {
+        /** @var Query $query */
+        $query = parent::query();
+        if (!Configure::read('debug')) {
+            return $query;
+        }
+
+        $traces = Debugger::trace(['start' => 2, 'depth' => 3, 'format' => 'array']);
+        $file = null;
+        $line = null;
+
+        foreach ($traces as $trace) {
+            $fullPath = $trace['file'];
+            $file = Debugger::trimPath($trace['file']);
+            $line = $trace['line'];
+            if (defined('CAKE_CORE_INCLUDE_PATH') && strpos($fullPath, CAKE_CORE_INCLUDE_PATH) !== 0) {
+                break;
+            }
+        }
+
+        $comment = sprintf('/* %s (line %s) */', $file, $line);
+        $query->epilog($query->newExpr($comment));
+
+        return $query;
+    }
+}
--- a/src/Model/Table/SqlTraceTrait.php
+++ b/src/Model/Table/SqlTraceTrait.php
@@ -13,8 +13,7 @@
  */
 namespace DebugKit\Model\Table;
 
-use Cake\Core\Configure;
-use Cake\Error\Debugger;
+use DebugKit\DebugSqlTemp;
 
 /**
  * Add this trait to your Table class to append the file reference of where a Query object was created.
@@ -30,25 +29,6 @@ trait SqlTraceTrait
      */
     public function query()
     {
-        $query = parent::query();
-
-        if (!Configure::read('debug') || $query->clause('epilog') !== null) {
-            return $query;
-        }
-
-        $traces = Debugger::trace(['start' => 2, 'depth' => 3, 'format' => 'array']);
-        $file = 'n/a';
-        $line = 0;
-
-        foreach ($traces as $trace) {
-            $path = $trace['file'];
-            $line = $trace['line'];
-            $file = Debugger::trimPath($path);
-            if (defined('CAKE_CORE_INCLUDE_PATH') && strpos($path, CAKE_CORE_INCLUDE_PATH) !== 0) {
-                break;
-            }
-        }
-
-        return $query->epilog($query->newExpr(sprintf('/* %s (line %s) */', $file, $line)));
+        return DebugSqlTemp::fileStamp(parent::query(), 2);
     }
 }
--- a/src/Model/Table/SqlTraceTrait.php
+++ b/src/Model/Table/SqlTraceTrait.php
@@ -30,26 +30,24 @@ trait SqlTraceTrait
     public function query()
     {
         $query = parent::query();
-        if (!Configure::read('debug')) {
+
+        if (!Configure::read('debug') || $query->clause('epilog') !== null) {
             return $query;
         }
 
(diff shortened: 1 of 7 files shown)


src/Model/Table/SqlTraceTrait.php
```diff
@@ -0,0 +1,56 @@
...
+ * @mixin \Cake\ORM\Table
+ */
+trait SqlTraceTrait
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function query()
+    {
+        /** @var Query $query */
```
dereuromark: Query could be FQCN inside the doc block

dereuromark:
:+1:

src/Model/Table/SqlTraceTrait.php
```diff
@@ -0,0 +1,55 @@
...
+            $fullPath = $trace['file'];
+            $file = Debugger::trimPath($trace['file']);
+            $line = $trace['line'];
+            if (defined('CAKE_CORE_INCLUDE_PATH') && strpos($fullPath, CAKE_CORE_INCLUDE_PATH) !== 0) {
+                break;
+            }
+        }
+
+        $comment = sprintf('/* %s (line %s) */', $file, $line);
+        $query->epilog($query->newExpr($comment));
```
markstory: You might want to append to the epilog instead of overwriting it. Or only appending if the epilog is empty. Right now you could corrupt queries if someone has also defined query() on their table.

thinkingmedia: @markstory Good point. I'll fix that.

src/Model/Table/SqlTraceTrait.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Creates a new Query instance for this repository
+     *
+     * @return \Cake\ORM\Query
+     */
+    public function query()
+    {
+        $query = parent::query();
+
+        if (!Configure::read('debug') || $query->clause('epilog') !== null) {
```
lorenzo: The epilog can actually be concatenated, no need to return early here

thinkingmedia: @lorenzo okay, I'm not sure how to do that because it could be a generic expression object not a string.

src/Model/Table/SqlTraceTrait.php
```diff
@@ -0,0 +1,54 @@
...
+        foreach ($traces as $trace) {
+            $path = $trace['file'];
+            $line = $trace['line'];
+            $file = Debugger::trimPath($path);
+            if (defined('CAKE_CORE_INCLUDE_PATH') && strpos($path, CAKE_CORE_INCLUDE_PATH) !== 0) {
+                break;
+            }
+        }
+
+        return $query->epilog($query->newExpr(sprintf('/* %s (line %s) */', $file, $line)));
```
lorenzo: I personally use `modifier()` instead of epilog for the same usage. `modifier` also has the advantage that it  automatically concatenates. It looks like

```sql
SELECT /* ... */
field1, field2...
```

Having it at the top is also great when you look at the currently running queries directly in your database

thinkingmedia: Thanks, I'll give it a try. I like the comment at the bottom (feels out of the way), but I'll give modifier a try (auto concating is worry free).

thinkingmedia:
I'm finished. Just want to wait on the other PR and then I'll rebase these changes and use `DebugSql` instead of `DebugSqlTemp`.

tests/TestCase/DebugSqlTempTest.php
```diff
@@ -0,0 +1,123 @@
...
+    public function testFileStampClosure()
+    {
+        $query = $this->newQuery()->select(['id']);
+        $func = function ($query) {
+            $this->assertSame($query, DebugSqlTemp::fileStamp($query));
+
+            return $query;
+        };
+        $query = $func($query);
+        // missing backslash on ROOT is an issue with test bootstrap, and not a bug
```
markstory: Shouldn't we fix the incorrect constant?

tests/TestCase/DebugSqlTempTest.php
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public function testFileStampDebugOff()
+    {
+        $query = $this->newQuery()->select(['id']);
+        $debug = Configure::read('debug');
+        Configure::write('debug', false);
+        $sql = (string)$query;
+        $this->assertSame($query, DebugSqlTemp::fileStamp($query, 1, true));
+        $this->assertEquals($sql, (string)$query);
+        Configure::write('debug', $debug);
```
markstory: The test harness should take care of restoring Configure state.

thinkingmedia: @markstory So there appears to be a bug in TestCase. On line 100 and it looks like this.

```
        if (!$this->_configure) {
            $this->_configure = Configure::read();
        }
```

The big is that `protected $_configure = []` has a default assignment which is truthy. That's breaking my tests and I have to manual restore the configuration. I'm sure this should have been an issue with other projects. I wonder why is hasn't been found before this?

tests/TestCase/DebugSqlTempTest.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * No comment should be set when debug is off.
+     * @skip
+     */
+    public function testFileStampDebugOff()
+    {
+        $query = $this->newQuery()->select(['id']);
+        // @todo Remove this when TestCase restoring of config is fixed.
```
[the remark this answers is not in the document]
thinkingmedia: oh come on. Seriously this is a stickler issue? That's exactly what a todo is for.

lorenzo:
Only one mode would be deseriable

thinkingmedia:
@lorenzo I removed the config option. It will just use modifier from now on.

tests/TestCase/DebugSqlTempTest.php
markstory: I'm not sure either. Where is the truthy value coming from? Empty lists are falsey in PHP

```php
php > if ([]) {
php {  echo 'yes';
php { } else {
php { echo 'no';
php { }
no
```

lorenzo:
@thinkingmedia sorry for the delay, do you mind resolving the conflicts so this can be merged?

ADmad:
@thinkingmedia Whatever strategy you follow for reminders is broken :stuck_out_tongue:.

dereuromark:
definitely broken :)
Shall we still finalize and merge this?

othercorey:
Should we close this? It's been a couple years since the last update.

PR #582, opened 2018-02-07
Put I18n to some strings

allanmcarvalho:
Some strings do not have a function, so it was added.

### DIFF
--- a/src/Controller/Component/ToolbarComponent.php
+++ b/src/Controller/Component/ToolbarComponent.php
@@ -39,7 +39,7 @@ class ToolbarComponent extends Component
     {
         parent::__construct($registry, $settings);
 
-        $msg = __d('debug_kit', "{0} is now loaded through plugin bootstrapping. Make sure you have {1} in your application's bootstrap.php.", 'DebugKit', '`Plugin::load("DebugKit", ["bootstrap" => true]);`');
+        $msg = __d('debug_kit', "{0} is now loaded through plugin bootstrapping. Make sure you have {1} in your application's {2}.", 'DebugKit', '`Plugin::load("DebugKit", ["bootstrap" => true]);`', 'bootstrap.php');
         throw new \RuntimeException($msg);
     }
 }
--- a/src/Controller/Component/ToolbarComponent.php
+++ b/src/Controller/Component/ToolbarComponent.php
@@ -39,8 +39,7 @@ class ToolbarComponent extends Component
     {
         parent::__construct($registry, $settings);
 
-        $msg = 'DebugKit is now loaded through plugin bootstrapping. Make sure you have ' .
-            '`Plugin::load("DebugKit", ["bootstrap" => true]);` in your application\'s bootstrap.php.';
+        $msg = __d('debug_kit', 'DebugKit is now loaded through plugin bootstrapping. Make sure you have `Plugin::load("DebugKit", ["bootstrap" => true]);` in your application\'s bootstrap.php.');
         throw new \RuntimeException($msg);
     }
 }
--- a/src/Controller/Component/ToolbarComponent.php
+++ b/src/Controller/Component/ToolbarComponent.php
@@ -39,7 +39,7 @@ class ToolbarComponent extends Component
     {
         parent::__construct($registry, $settings);
 
-        $msg = __d('debug_kit', 'DebugKit is now loaded through plugin bootstrapping. Make sure you have `Plugin::load("DebugKit", ["bootstrap" => true]);` in your application\'s bootstrap.php.');
+        $msg = __d('debug_kit', "{0} is now loaded through plugin bootstrapping. Make sure you have {1} in your application's bootstrap.php.", 'DebugKit', '`Plugin::load("DebugKit", ["bootstrap" => true]);`');
         throw new \RuntimeException($msg);
     }
 }
--- a/src/Model/Table/PanelsTable.php
+++ b/src/Model/Table/PanelsTable.php
@@ -56,7 +56,7 @@ class PanelsTable extends Table
     public function findByRequest(Query $query, array $options)
     {
         if (empty($options['requestId'])) {
-            throw new \RuntimeException('Missing request id in findByRequest.');
+            throw new \RuntimeException(__d('debug_kit', 'Missing request id in findByRequest.'));
         }
 
         return $query->where(['Panels.request_id' => $options['requestId']])
--- a/src/Model/Table/PanelsTable.php
+++ b/src/Model/Table/PanelsTable.php
@@ -56,7 +56,7 @@ class PanelsTable extends Table
     public function findByRequest(Query $query, array $options)
     {
         if (empty($options['requestId'])) {
-            throw new \RuntimeException(__d('debug_kit', 'Missing request id in findByRequest.'));
+            throw new \RuntimeException(__d('debug_kit', 'Missing request id in {0}.', 'findByRequest()'));
         }
 
         return $query->where(['Panels.request_id' => $options['requestId']])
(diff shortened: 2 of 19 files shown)


src/Template/Element/cache_panel.ctp
```diff
@@ -69,9 +69,9 @@ $(document).ready(function() {
             type: 'POST'
         });
         xhr.done(function(response) {
-            showMessage(messageEl, name + ' cache cleared.')
+            showMessage(messageEl, name + ' ' . __d('debug_kit', 'cache cleared.'))
         }).error(function(response) {
-            showMessage(messageEl, name + ' cache could not be cleared.');
+            showMessage(messageEl, name + ' ' . __d('debug_kit', 'cache could not be cleared.'));
```
markstory: I don't think you're in PHP code here. You need the `<?= ?>` tags.

src/Controller/Component/ToolbarComponent.php
```diff
@@ -39,8 +39,7 @@ public function __construct(ComponentRegistry $registry, $settings = [])
     {
         parent::__construct($registry, $settings);
 
-        $msg = 'DebugKit is now loaded through plugin bootstrapping. Make sure you have ' .
-            '`Plugin::load("DebugKit", ["bootstrap" => true]);` in your application\'s bootstrap.php.';
+        $msg = __d('debug_kit', 'DebugKit is now loaded through plugin bootstrapping. Make sure you have `Plugin::load("DebugKit", ["bootstrap" => true]);` in your application\'s bootstrap.php.');
```
dereuromark: This is a really bad idea.
If you start using translations, never ever do that with code pieces or actual literal words that must not be translated, that includes the plugin name here, as well as the the complete code piece.

So use the {0} {1} ... placeholders then.

For the record: I still think it would be wiser to completly remove the translations.

src/Model/Table/PanelsTable.php
```diff
@@ -56,7 +56,7 @@ public function initialize(array $config)
     public function findByRequest(Query $query, array $options)
     {
         if (empty($options['requestId'])) {
-            throw new \RuntimeException('Missing request id in findByRequest.');
+            throw new \RuntimeException(__d('debug_kit', 'Missing request id in findByRequest.'));
```
dereuromark: findByRequest should be findByRequest()
and it should not be part of the translation string!

src/Shell/WhitespaceShell.php
```diff
@@ -95,7 +95,7 @@ public function getOptionParser()
 
         return $parser->addOption('path', [
             'short' => 'p',
-            'help' => __d('cake_console', 'Absolute path or relative to APP.')
+            'help' => __d('debug_kit', 'Absolute path or relative to APP.')
```
dereuromark: APP must not be part of translation string

src/Template/Element/mail_panel.ctp
```diff
@@ -1,13 +1,13 @@
 <p class="info">
-    Why not test your emails interactively instead? Go to the
-    <?= $this->Html->link('Email previews page',
+    <?= __d('debug_kit', 'Why not test your emails interactively instead? Go to the') ?>
```
dereuromark: "Go to the" makes no sense.
It would have to be "Go to the {0}" with the HTML URL injected.

src/Template/MailPreview/index.ctp
```diff
@@ -26,14 +26,14 @@
 <?php endforeach; ?>
 
 <div style="font-weight:300;margin-top:100px">
-    <h3>How to use this feature?</h3>
-    <p>Testing emails can be very time consuming &#8987;</p>
-    <p> Specially when you need to click a bunch of times on an interface to trigger them.</p>
-    <p>Wouldn't it be better to just change the templates and refresh the browser to see the result?</p>
-    <p>Just the way you work on the web interface! &#127939;</p>
+    <h3><?= __d('debug_kit', 'How to use this feature?') ?></h3>
```
dereuromark: So much overhead for so little benefit for anyone, in the future :/

src/Template/MailPreview/index.ctp
```diff
@@ -58,7 +58,7 @@
         highlight_string($code);
     ?>
     </pre>
-    <p>Now you create a MailPreview class where you can pass some dummy values.</p>
+    <p><?= __d('debug_kit', 'Now you create a MailPreview class where you can pass some dummy values.') ?></p>
```
dereuromark: MailPreview must not be part of translation string, it is a class name.

src/Template/MailPreview/index.ctp
```diff
@@ -84,7 +84,7 @@
     ?>
     </pre>
 
-    <p>Note that the function MUST return the UserMailer object at the end.</p>
-    <p>Since Mailers have a fluent interface, you just need to return the result of the chain of calls.</p>
-    <p style="margin:20px 0">That's it, now refresh this page! &#128579;</p>
+    <p><?= __d('debug_kit', 'Note that the function MUST return the UserMailer object at the end.') ?></p>
```
dereuromark: UserMailer must not be part of translation string, it is a class name.

dereuromark:
For the record: 👎 as the costs are too high for the theoretical (and practical 0 use case) benefit.
I still think it would be wiser to completly remove the translations.

allanmcarvalho:
@dereuromark Well, I believe the code should maintain a homogeneity. Yours arguments are somewhat valid. Although I believe that having the translation would be beneficial at some time and for someone, but I also agree that making this translation work fully costs will be high. Whatever you decide, as long as you keep a pattern (all with, or all without), I will somehow find that the best has been done.

markstory:
@dereuromark How is there 0 practical use case? Once all the UI strings are in `__()` we could have contributors send translations for DebugKit, allowing them to see the content in their native language.

src/Controller/Component/ToolbarComponent.php
```diff
@@ -39,7 +39,7 @@ public function __construct(ComponentRegistry $registry, $settings = [])
     {
         parent::__construct($registry, $settings);
 
-        $msg = __d('debug_kit', 'DebugKit is now loaded through plugin bootstrapping. Make sure you have `Plugin::load("DebugKit", ["bootstrap" => true]);` in your application\'s bootstrap.php.');
+        $msg = __d('debug_kit', "{0} is now loaded through plugin bootstrapping. Make sure you have {1} in your application's bootstrap.php.", 'DebugKit', '`Plugin::load("DebugKit", ["bootstrap" => true]);`');
```
dereuromark: `bootstrap.php` is also a code piece so to speak

src/Template/MailPreview/index.ctp
```diff
@@ -47,18 +47,18 @@
         {
             public function welcome($user)
             {
-                return $this // Returning the chain is a good idea :)
+                return $this // ' . __d('debug_kit', 'Returning the chain is a good idea') . ' :)
```
dereuromark: These are comments, there is no code here.

src/Template/MailPreview/index.ctp
```diff
@@ -47,18 +47,18 @@
...
                     ->to($user->email)
                     ->subject(sprintf("Welcome %s", $user->name))
-                    ->template("welcome_mail") // By default template with same name as method name is used.
+                    ->template("welcome_mail") // ' . __d('debug_kit', 'By default template with same name as method name is used.') . '
```
dereuromark: Same

src/Template/MailPreview/index.ctp
allanmcarvalho: I know.... But this comments are showed in UI

dereuromark: They are? But then they will not show translated with this here either, as this is not valid PHP syntax.

allanmcarvalho: Yes, it does, because this sample code is inside a PHP '$code' variable. That is, it is not HTML but a PHP example inside a variable. So I concatenated the function I18N with the variable string content

Issue #554, opened 2017-09-29
Feature: Add dashboard


dereuromark:
This is what I usually do for my plugins, and it really helps to get an overview of the functionality provided and that one doesnt have to remember specific URLs or find them in the toolbar at the bottom.

So Dashboard URL entrypoint would be `/debug-kit` directly.
It contains (ideas):

- [x] Size of current DB and button to clear/reset
- [x] Link to MailPreview

We can also then piece by piece add more stuff:
- [ ] More statistics and metrics.
- [ ] current PHP and DB versions, etc.
- [ ] Last requests to inspect in detail maybe
- [ ] Other useful dashboard toolings that can be provided that make sense to have there (more global than on the bottom of a specific page as toolbar).

markstory:
Sounds like a neat idea. This might also be a good place to explore more history as well.

dereuromark:
The other enhancements can be follow PRs. I will close this now.

PR #593, opened 2018-03-02
Add basic dashboard.

dereuromark:
First two points of https://github.com/cakephp/debug_kit/issues/554

This can probably be improved in steps.


markstory:
Where this be attached to the toolbar at all?

dereuromark:
I didnt yet, but we can.
I mainly thought people could link this in their admin backend (as I always do) where needed.
But it seems useful to also have a buton in the toolbar.
And then even move "global" things into there instead of having so many tabs at the bottom?

markstory:
>  And then even move "global" things into there instead of having so many tabs at the bottom?

That could work. The toolbar should have all the information for the current request, but cache clearing, database reset, and other visualizations could live in a dashboard.

lorenzo:
Do you mind attaching a screenshot of this dashboard?

ADmad:
@dereuromark Ping

dereuromark:
Is someone interested in finishing this up?
As I mentioned this would be an incremental thing probably.

dereuromark:
debugKit

here you go.
I think for a first iteration this is probably fair enough.

src/Controller/DebugKitController.php
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @return \Cake\Http\Response|null
+     */
+    public function reset()
+    {
+        $this->request->allowMethod('post');
+
+        $connection = ConnectionManager::getConfig('debug_kit');
+        if ($connection['driver'] === Sqlite::class && file_exists($connection['database'])) {
+            unlink($connection['database']);
```
markstory: Instead of deleting the file could we truncate the tables? That would let this method be used for other databases.

src/Controller/DebugKitController.php
```diff
@@ -0,0 +1,81 @@
...
+     * Dashboard.
+     *
+     * @return void
+     * @throws \Cake\Network\Exception\NotFoundException
+     */
+    public function index()
+    {
+        $connection = ConnectionManager::getConfig('debug_kit');
+
+        if ($connection['driver'] === Sqlite::class) {
```
markstory: For other databases we could list the database name + row count.

src/Controller/DebugKitController.php
dereuromark: I am sure we could.

dereuromark:
> Syntax error or access violation: 1235 This version of MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery'

PR #594, opened 2018-03-03
Remove deprecations to work in 3.6.

dereuromark:
The following is needed for 3.6 to work due to the deprecation notices being thrown.


saeideng:
@dereuromark  I re-based to 3.next branch 
and with solving conflicts your commit lost , sorry

dereuromark:
Maybe all my changes were already in then? I didnt know about the 3.next branch and just applied the changes to master directly.
So maybe we can close then.

saeideng:
>Maybe all my changes were already in then?

yes I checked

PR #595, opened 2018-03-04
Prevent Debugkit from running in production

lorenzo:
Running an application with debug mode on AND installing dev dependencies in a
production environment is obviously the developers' fault for not following the
instructions provided. But we can do better than that.

Instead of blaming them when their secrets are exposed to the internet, we can
run some simple checks that can tells us whether or not the aplications is running
visibly to the www.

In the valid cases developers need DebugKit running in producition-like systems (like a
QA environment), they still have the option of `forceEnable`.

### DIFF
--- a/src/ToolbarService.php
+++ b/src/ToolbarService.php
@@ -94,15 +94,43 @@ class ToolbarService
     public function isEnabled()
     {
         $enabled = (bool)Configure::read('debug');
-        if ($enabled) {
+
+        if ($enabled && !$this->isNotSuspiciouslyProduction()) {
             return true;
         }
+
         $force = $this->config('forceEnable');
         if (is_callable($force)) {
             return $force();
         }
 
         return $force;
+    }
+
+    /**
+     * Returns true if this applications is being executed on a domain with a TLD
+     * that is commonly associated with a production environment.
+     *
+     * @return bool
+     */
+    protected function isNotSuspiciouslyProduction()
+    {
+        $host = explode('.', env('HTTP_HOST'));
+        $first = current($host);
+        $isIP = is_numeric(implode('', $host));
+
+        if (count($host) === 1) {
+            return false;
+        }
+
+        if ($isIP && in_array($first, ['192', '10'])) {
+            // Accessing the app by private IP, this is safe
+            return false;
+        }
+
+        $tld = end($host);
+        $safeTLD = ["localhost", "dev", "invalid", "test", "example"];
+        return !in_array($tld, $safeTLD);
     }
 
     /**
--- a/tests/TestCase/ToolbarServiceTest.php
+++ b/tests/TestCase/ToolbarServiceTest.php
@@ -61,6 +61,17 @@ class ToolbarServiceTest extends TestCase
     }
 
     /**
+     * teardown
+     *
+     * @return void
+     */
+    public function tearDown()
+    {
+        parent::tearDown();
+        putenv('HTTP_HOST=');
+    }
+
+    /**
      * Test loading panels.
      *
      * @return void
@@ -342,6 +353,41 @@ class ToolbarServiceTest extends TestCase
         Configure::write('debug', false);
         $bar = new ToolbarService($this->events, []);
         $this->assertFalse($bar->isEnabled(), 'debug is off, panel is disabled');
+    }
+
+    /**
+     * Test isEnabled returns false for some suspiciously production environments
+     *
+     * @param string $domain The domain name where the app is hosted
+     * @param bool $isEnabled The expectation for isEnabled()
+     * @dataProvider domainsProvider
+     * @return void
+     */
+    public function testIsEnabledProductionEnv($domain, $isEnabled)
+    {
+        Configure::write('debug', true);
+        putenv("HTTP_HOST=$domain");
+        $bar = new ToolbarService($this->events, []);
+        $this->assertEquals($isEnabled, $bar->isEnabled());
+    }
+
+    public function domainsProvider()
+    {
+        return [
+            ['localhost', true],
+            ['192.168.1.34', true],
+            ['10.14.34.5', true],
+            ['myapp.localhost', true],
+            ['myapp.dev', true],
+            ['myapp', true],
+            ['myapp.invalid', true],
+            ['myapp.test', true],
+            ['myapp.com', false],
+            ['myapp.io', false],
+            ['myapp.net', false],
+            ['172.112.34.2', false],
+            ['6.112.34.2', false],
+        ];
     }
 
     /**


src/ToolbarService.php
```diff
@@ -105,6 +107,33 @@ public function isEnabled()
...
+            return false;
+        }
+
+        if ($isIP && in_array($first, ['192', '10'])) {
+            // Accessing the app by private IP, this is safe
+            return false;
+        }
+
+        $tld = end($host);
+        $safeTLD = ["localhost", "dev", "invalid", "test", "example"];
```
dereuromark: .local missing IMO.

lorenzo: good idea

thinkingmedia:
I've never in my many years of using CakePHP ever deployed to production an application that had debugKit turned on. I also don't connect my database to the internet or use "password" for my root user account.

The problem I have with this feature is that it hides the fact that the developer made a huge security mistake. I don't mean a small mistake but a really big one like they shared their entire database schema, variables and other such things. It is a stupid thing to do.

People who do that. They deserve the consequences of those mistakes, because they are not taking the time to read the documentation, follow basic practices and put in the little bit of effort to protect their production environment. Those developers won't be CakePHP developers for very long.

As a result of that, the rest of us CakePHP developers don't make this mistake.

Which raises the question of how helpful or harmful this change will be.

It is far more likely an experienced CakePHP developer is going to be scratching his head as to *why* the debugKit disappeared on his staging and testing servers. We know this person is professional because they configured a DNS properly for staging and don't use things like localhost or IP addresses.

There will be more CakePHP developers who see their debugKits disappear magically and have to waste time googling how to bring it back, than there will be developers deploying to production with their debugKits turned on and even if there are developers doing that. They shouldn't be doing that and we shouldn't be helping them this way because they aren't learning an important lesson about trying to be a better developer.

thinkingmedia:
Instead of disabling the debugKit why not replace the debugKit icon with an exclamation mark, and on the mouse over event display a message that says "debugKit is disabled in production, set forceEnable to true to enable production mode".

This solves someone deploying to production by mistake, tells them what they did wrong and if this is really what they wanted a way to enable debugKit anyway.

I apologize for my previous comment if it was offensive.

dereuromark:
I also see more the need to "more in your face" output some warning, throw exceptions even, rather than silently hiding their mistakes.

ADmad:
@thinkingmedia I am of the same opinion that having debugkit enabled in production is the developer's fault but thing is as framework devs we are expected to safeguard people from shooting themselves in the foot. Having debugkit (and app secrets as a consequence) show up in production is bad publicity for us irrespective of who's fault it.

I also agree that silently hiding debugkit is not a good idea. Either showing an icon with a message like you suggested or simply throwing an exception as @dereuromark suggested would be a better idea.

lorenzo:
I can add a log message for when debug kit was disabled due to the domain check.

My idea was to also change the app template to make visible the option to force enable debug kit if needed.

lorenzo:
@thinkingmedia We all make mistakes. Sometimes those mistakes are shipped to production, regardless of how savvy we are of production deploys.

thinkingmedia:
@lorenzo how about a DNS whitelist method of enabling the debugKit. The developer has to add the DNS to an array otherwise it's off by default. This way it forces people to define which DNS entries are granted debugKit access rather than it being dependant upon a boolean flag.

lorenzo:
@thinkingmedia that's what the `foreEnable` feature is for. I'm going to add some sample code in the app template showing how to enable for certain domains

lorenzo:
@thinkingmedia I think the detail that you are missing here is that `forceEnable` can be a callable. You can check things like the current domain there, or certain get arguments, for example

src/ToolbarService.php
```diff
@@ -94,9 +94,11 @@ public function registry()
     public function isEnabled()
     {
         $enabled = (bool)Configure::read('debug');
-        if ($enabled) {
+
+        if ($enabled && !$this->isNotSuspiciouslyProduction()) {
```
markstory: :clap: great method name.

src/ToolbarService.php
```diff
@@ -105,6 +107,33 @@ public function isEnabled()
...
+    {
+        $host = explode('.', parse_url('http://' . env('HTTP_HOST'), PHP_URL_HOST));
+        $first = current($host);
+        $isIP = is_numeric(implode('', $host));
+
+        if (count($host) === 1) {
+            return false;
+        }
+
+        if ($isIP && in_array($first, ['192', '10', '127'])) {
```
markstory: My production environment at work has webservers on 10. addresses, as our public network terminates at the load balancer in front of the webservers. If we want to err on the cautious side we might not want to use this check.

dereuromark:
Guys, the DebugKit thing (and display/hide) is only the symptom of the decease, not the root issue at hand.
The actual problem is that debug mode is still on, and even with hiding the toolbar, the main concern here is the verbose error and exception output. That will still persist then - and is way more critical IMO.
It means that anyone can read the complete strack trace including calling arguments which very likely contain still the plain text credentials and stuff.

So while this might hide a few non-critical elements being displayed, the critical security issue of debug on forever is factor 100 higher.

I can only repeat @thinkingmedia in how important it is to educate and school people in this very basic common sense thing of never ever leaving debug mode on in production.
And that by design will make this PR somewhat not necessary.

PS: We can still use forceEnable to enable the toolbar, even with debug mode off. Ideally that is done in a session or IP way so not everyone gets it enabled, but here debug mode doesnt have to be on in the first place either :)

src/ToolbarService.php
lorenzo: @markstory but the request host would not be the IP unless you had access to the local private network, that's why I think it is safe.

lorenzo:
@dereuromark I agree that education is key. At least with this change the potential for disaster is less than before.

dereuromark:
Dunno, as I said: The stack tracing on (deliberate) exceptions caused due to the open debug mode is usually way more catastrophic then the already somewhat sanitized debug kit output :)

saeideng:
Can you move to 3.next?
And maybe with a bit note in app/3.next

src/ToolbarService.php
markstory: Good point.

lorenzo:
I merged 3.next in this branch, and retargeted to 3.next. That brought a bunch of other commits here, but at least the branch is up to date

davidyell:
This feature is missing documentation.

lorenzo:
yeah, haven't found a couple minutes to add this to the docs, sorry @davidyell

davidyell:
For some reason, even after configuring the `forceEnable`, DebugKit is logging 10 force disables per request to my debug log.

https://i.imgur.com/xoG0rEg.png

dereuromark:
We should revert this. I said already above and in chat a few times quite clearly that this is a bad idea and compromises actually security way more. It cloaks the debug mode being on in a way that stack traces reveal way more dangerous info. Instead people should be more aware to disable debug in prod.

lorenzo:
@davidyell can you paste that on a separate ticket so it can be tracked and fixed?

lorenzo:
@davidyell I would also love to know the TLD for the domain you are running debugkit. Did we miss one important case?

davidyell:
@lorenzo Our TLDs for development are set by our internal DNS, and each machine has it's own number. Which means my machine is `ukem157`, but my colleagues machine is `ukwm500`.

PR #596, opened 2018-03-04
replace with undeprecated methods

saeideng:
https://github.com/cakephp/debug_kit/pull/594#issuecomment-370168677

### DIFF
--- a/src/Panel/SqlLogPanel.php
+++ b/src/Panel/SqlLogPanel.php
@@ -14,8 +14,8 @@ namespace DebugKit\Panel;
 
 use Cake\Core\Configure;
 use Cake\Datasource\ConnectionManager;
+use Cake\ORM\Locator\TableLocator;
 use Cake\ORM\Table;
-use Cake\ORM\TableRegistry;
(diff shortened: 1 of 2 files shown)


src/Panel/SqlLogPanel.php
```diff
@@ -79,7 +79,7 @@ public function data()
         return [
             'tables' => array_map(function (Table $table) {
                 return $table->getAlias();
-            }, TableRegistry::genericInstances()),
+            }, (new TableLocator)->genericInstances()),
```
lorenzo: You need the global instance for this feature to work correctly

saeideng: use of `LocatorAwareTrait` ? maybe

lorenzo: Yes!

saeideng: `LocatorAwareTrait::getTableLocator()` uses `TableRegistry::getTableLocator();` internally 
I changed to use `TableRegistry::getTableLocator();`

PR #598, opened 2018-03-06
Improve Includes panel

garas:
* Plugins were uncategorized on Windows because \ and / mismatch.
* Include panel missed many included files if they are not app/cake/plugin - add uncategorized files to 'other' group
* Many includes are from Composer vendor/ directory - add 'vendor' category and group by package
* Include panel summary displayed incorrect count, because it included count of 'Include Paths'
* Changed subgroup of app/cake/plugin to lowercase 'other' to be distinct from Controller/Model/... and to be last in list when sorted.

I've tried to add test like `$this->assertArrayHasKey('cakephp/chronos', $data['vendor']);`, but during test everything (cake/vendor/etc) is inside DebugKit directory, so everything goes into 'plugins -> DebugKit' group.

But in normal app setup, everything goes into correct categories.

### DIFF
--- a/src/Panel/IncludePanel.php
+++ b/src/Panel/IncludePanel.php
@@ -71,6 +71,8 @@ class IncludePanel extends DebugPanel
                 $return['app'][$this->_getFileType($file)][] = $this->_niceFileName($file, 'app');
             } elseif ($this->_isCakeFile($file)) {
                 $return['cake'][$this->_getFileType($file)][] = $this->_niceFileName($file, 'cake');
+            } else {
+                $return['other'][] = $this->_niceFileName($file, 'root');
             }
         }
 
@@ -137,7 +139,7 @@ class IncludePanel extends DebugPanel
     }
 
     /**
-     * Replace the path with APP, CORE or the plugin name
+     * Replace the path with APP, CAKE, ROOT or the plugin name
      *
      * @param string $file File to check
      * @param string $type The file type
@@ -154,6 +156,9 @@ class IncludePanel extends DebugPanel
 
             case 'cake':
                 return str_replace(CAKE, 'CAKE' . DIRECTORY_SEPARATOR, $file);
+
+            case 'root':
+                return str_replace(ROOT, 'ROOT', $file);
 
             default:
                 return str_replace($this->_pluginPaths[$type], $type . DIRECTORY_SEPARATOR, $file);
--- a/src/Panel/IncludePanel.php
+++ b/src/Panel/IncludePanel.php
@@ -82,6 +82,10 @@ class IncludePanel extends DebugPanel
         ksort($return['plugins']);
         ksort($return['app']);
 
+        foreach ($return['plugins'] as &$plugin) {
+            ksort($plugin);
+        }
+
         return $return;
     }
 
--- a/src/Panel/IncludePanel.php
+++ b/src/Panel/IncludePanel.php
@@ -49,7 +49,7 @@ class IncludePanel extends DebugPanel
     public function __construct()
     {
         foreach (Plugin::loaded() as $plugin) {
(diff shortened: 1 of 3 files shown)


src/Panel/IncludePanel.php
```diff
@@ -137,16 +170,31 @@ protected function _isPluginFile($file)
...
+     * Check if a path is from a Composer package
+     *
+     * @param string $file File to check
+     * @return bool
+     */
+    protected function _isVendorFile($file)
+    {
+        foreach ($this->_composerPaths as $package => $path) {
+            if (strstr($file, $path)) {
+                return $package;
```
dereuromark: Is this of type bool?

src/Panel/IncludePanel.php
```diff
@@ -137,16 +170,31 @@ protected function _isPluginFile($file)
...
      * @param string $file File to check
      * @param string $type The file type
-     *  - app for app files
-     *  - cake for cake files
-     *  - PluginName for the name of a plugin
+     * @param string $name plugin name or composer package
      * @return bool
      */
-    protected function _niceFileName($file, $type)
+    protected function _niceFileName($file, $type, $name = null)
```
dereuromark: no need for name to be nullable

dereuromark:
Nice work!

src/Panel/IncludePanel.php
garas: `strstr` returns string, but I just copy/pasted code from `_isPluginFile` .

src/Panel/IncludePanel.php
garas: Function gets 2 arguments when `$type` is app/cake/root, `$name` is used when type is plugin/vendor.

dereuromark: Then you should fix the doc block.

src/Panel/IncludePanel.php
dereuromark: You return $package, but doc block states bool.

src/Panel/IncludePanel.php
```diff
@@ -31,6 +32,13 @@ class IncludePanel extends DebugPanel
      */
     protected $_pluginPaths = [];
 
+    /**
+     * The list of Composer packages
+     *
+     * @var <type>
```
dereuromark: type? array

src/Panel/IncludePanel.php
```diff
@@ -156,7 +156,7 @@ protected function _isAppFile($file)
      * Check if a path is from a plugin
      *
      * @param string $file File to check
-     * @return bool
+     * @return string|bool plugin name, or false if not plugin
```
dereuromark: I dont think an is...() method should ever return anything but bool. Why do you need to return the plugin name?
Then it should be called vendorFile() and return null otherwise.

src/Panel/IncludePanel.php
```diff
@@ -173,7 +173,7 @@ protected function _isPluginFile($file)
      * Check if a path is from a Composer package
      *
      * @param string $file File to check
-     * @return bool
+     * @return string|bool package name, or false if not Composer package
```
dereuromark: Same here.

markstory:
Once the tests are passing this looks pretty good to me.

garas:
Tests are failing because CakePHP 3.6 deprecations.

Tests on 3.4.0 (`PHP: 5.6 PREFER_LOWEST=1`) passes

markstory:
My mistake @garas. We have another branch that has no remaining deprecation errors, but it might be some time before that lands on master.

saeideng:
fixed in https://github.com/cakephp/debug_kit/pull/599
@garas  can you rebase your PR on master?

garas:
I've rebased yesterday, tests passed. You can merge.

PR #605, opened 2018-04-16
attempt to show deprecation warning in debug_kit

saeideng:
is not yet complete 
for testing 
```php
// /config/bootstrap.php

$isCli = PHP_SAPI === 'cli';
if ($isCli) {
    (new ConsoleErrorHandler(Configure::read('Error')))->register();
} else {

    (new \DebugKit\Error\ErrorHandler(Configure::read('Error')))->register();
    //(new ErrorHandler(Configure::read('Error')))->register();
}
```


src/Panel/DeprecatedPanel.php
```diff
@@ -0,0 +1,268 @@
...
+        if ($lockFile->exists()) {
+            $lockContent = $lockFile->read();
+
+            $vendorDir = ROOT . DIRECTORY_SEPARATOR . 'vendor' . DIRECTORY_SEPARATOR;
+            $packages = array_merge($lockContent['packages'], $lockContent['packages-dev']);
+
+            foreach ($packages as $package) {
+                $this->_composerPaths[$package['name']] = $vendorDir . str_replace('/', DIRECTORY_SEPARATOR, $package['name']) . DIRECTORY_SEPARATOR;
+            }
+        }
```
markstory: Should this be extracted into an object that can be used by both this and the include panel? Some sort of path classifier?

src/Panel/DeprecatedPanel.php
```diff
@@ -0,0 +1,268 @@
...
+     */
+    protected function _prepare()
+    {
+        $errors = \DebugKit\ToolbarService::getDeprecatedErrors();
+        $return = ['cake' => [], 'app' => [], 'plugins' => [], 'vendor' => [], 'other' => []];
+
+        foreach ($errors as $error) {
+            $array = explode(' ', $error['description']);
+            $line = $array[count($array) - 1];
+            $file = $array[count($array) - 3];
```
markstory: Couldn't the errors be appended with the file/line as part of the data structure? Munging strings seems like it could be error prone.

src/Template/Element/deprecated_panel.ctp
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+// Backwards compat for old DebugKit data.
+if (!isset($cake) && isset($core)) {
+    $cake = $core;
+}
+?>
+
+<h4><?= __d('debug_kit', 'Deprecated methods paths') ?></h4>
+<?= $this->Toolbar->makeNeatArray(compact('app', 'cake', 'plugins', 'vendor', 'other')) ?>
```
markstory: It might be easier to read if this information was presented as a table for each package/app. Right now you'd need a lot of clicking to quickly scan your deprecations.

src/Error/ErrorHandler.php
```diff
@@ -0,0 +1,57 @@
...
+use Cake\Error\ErrorHandler as BaseErrorHandler;
+use Cake\Http\ResponseEmitter;
+use DebugKit\ToolbarService;
+use Exception;
+use Throwable;
+
+/**
+ * @see \Cake\Error\ExceptionRenderer for more information
+ */
+class ErrorHandler extends BaseErrorHandler
```
markstory: Should the plugin register this error handler in its bootstrap hook? That would make the deprecation workflow with debugkit enabled easier.

src/Panel/DeprecatedPanel.php
saeideng: $error['file'] shows `core/function.php`

src/Error/ErrorHandler.php
saeideng: >Should the plugin register this error handler in its bootstrap hook?
just need to register in `app/config/bootstrap.php`

src/Error/ErrorHandler.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    protected function _displayError($error, $debug)
+    {
+        if (!$debug) {
+            return;
+        }
+        if (strtolower($error['error']) !== 'deprecated') {
+            parent::_displayError($error, $debug);
+        }
+        if (!Plugin::loaded('DebugKit')) {
```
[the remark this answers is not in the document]
saeideng: fixed

src/Panel/DeprecatedPanel.php
saeideng: fixed

src/Template/Element/deprecated_panel.ctp
saeideng: a bit changed

src/Error/ErrorHandler.php
saeideng: how we can enable it in `{plugin}config/bootstrap.php` ?

markstory: Instead of an ErrorHandler class we could do something like the following in debug kit's bootstrap:

```php
$previousHandler = set_error_handler(
    function ($code, $message, $file, $line) use (&$previousHandler) {
        if ($code == E_USER_DEPRECATED || $code == E_DEPRECATED) {
            DeprecatedPanel::addDeprecation($code, $message, $file, $line);
            return;
        }
        if ($previousHandler) {
            return $previousHandler($code, $message, $file, $line);
        }
    }
);
```

This approach would remove the deprecations from the logs and HTML display, and only display them in DebugKit.

saeideng:
for disable deprecation panel + turn off error handler 
```php
// config/bootstrap.php
if (Configure::read('debug')) {
    Configure::write('DebugKit.panels', ['DebugKit.Deprecated' => false]);
    Plugin::load('DebugKit', ['bootstrap' => true']);
}
```

saeideng:
@markstory  finished.

src/IncludePanel.php
```diff
@@ -0,0 +1,201 @@
...
+use Cake\Event\Event;
+use Cake\Utility\Hash;
+use Composer\Json\JsonFile;
+use DebugKit\DebugPanel;
+
+/**
+ * Provides a list of files for the current request
+ *
+ */
+class IncludePanel extends DebugPanel
```
markstory: Can this problem be solved with composition instead of inheritance? Having two classes named IncludePanel is awkward.

saeideng: done.

src/DebugInclude.php
```diff
@@ -0,0 +1,190 @@
...
+use Cake\Event\Event;
+use Cake\Utility\Hash;
+use Composer\Json\JsonFile;
+use DebugKit\DebugPanel;
+
+/**
+ * Contains methods for Providing list of files.
+ *
+ */
+class DebugInclude
```
markstory: Much nicer imo.

src/Template/Element/deprecated_panel.ctp
lorenzo: what's the decision on this?

saeideng: nothing yet,
I no have idea about this.

markstory:
What do you think of a list style output instead of the 'neat array'? I think this makes the deprecations easier to read as there is no need to click.

screen shot 2018-04-28 at 09 37 44

This only shows one section, but there would be separate sections for each top level group containing deprecations.

saeideng:
I'm OK with both 
just needs to change this line
```php
 $return['app'][$this->_debug->getFileType($file)][$this->_debug->niceFileName($file, 'app')][] = $description; 
```

> I think this makes the deprecations easier to read as there is no need to click.

I think only  `app` deprecations is enough for this change

markstory:
@saeideng I've pushed up some changes to make the output easier to read. The code is missing tests, but I can add those once we're happy with the output.

saeideng:
looks good , :+1: 
just I receive 
```php
Vendor

    Notice (8): Undefined index: niceFile [/var/.../src/Template/Element/deprecations_panel.ctp, line 35]

    :

    Notice (8): Undefined index: line [/var/.../src/Template/Element/deprecations_panel.ctp, line 35]


    Notice (8): Undefined index: message [/var/.../src/Template/Element/deprecations_panel.ctp, line 37]
```

saeideng:
:+1:
new deprecation errors fixed in other PR and unrelated to this PR

PR #607, opened 2018-04-17
[2.2] Fix explainQuery not working with PHP7.2 

chinpei215:
I have changed Object to Model (with useTable = false). Though I am not sure I can change the class tree in a bugfix release, direct use of CakeObject would break legacy 2.x applications.
Instead I could use `class_alias('Object', 'CakeObject')` if the `CakeObject` class was not found. But if users or some plugin's author also did it withtout class_exists() checks, those applications might break. So I thought that using Model class would be safe. But if you prefer class_alias(), I will revise my commits. Either way is fine for me if it works with PHP 7.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: precise
+
 php:
   - 5.3
   - 5.4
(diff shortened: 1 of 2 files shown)


markstory:
This is fine by me.

PR #619, opened 2018-05-04
refactoring for CakePHP 3.6

mrothauer:


mrothauer:
sorry, needs to be pulled against 3.next

PR #620, opened 2018-05-04
refactoring for CakePHP 3.6

mrothauer:


src/Controller/ToolbarController.php
```diff
@@ -61,11 +61,11 @@ public function beforeFilter(Event $event)
      */
     public function clearCache()
     {
-        $this->request->allowMethod('post');
-        if (!$this->request->getData('name')) {
+        $this->getRequest()->allowMethod('post');
+        if (!$this->getRequest()->getData('name')) {
```
markstory: Controller's don't need to use `getRequest()` as the request is a protected property.

src/Controller/MailPreviewController.php
```diff
@@ -128,7 +128,7 @@ public function email($name, $method)
         $this->set('title', $humanName);
         $this->set('email', $email);
         $this->set('plugin', $plugin);
-        $this->set('part', $this->findPreferredPart($email, $this->request->getQuery('part')));
+        $this->set('part', $this->findPreferredPart($email, $this->getRequest()->getQuery('part')));
```
markstory: Controller's don't need to use `getRequest()` as the request is a protected property.

src/Controller/ToolbarController.php
mrothauer: in the 3.next branch of cakephp/cakephp the $request property is public and deprecated

https://github.com/cakephp/cakephp/blob/c40ab0a49953ff6bd8a8397b35869e229be00a26/src/Controller/Controller.php#L128-L132

ADmad: @mrothauer The deprecation is in regards to it's visibility. It's going to be made protected instead of public, not removed. So a controller internally can still use `$this->request`.

mrothauer: Now I got it, it's written clearly: "The property will become protected in 4.0.0." Sorry.

markstory:
Thank you 👍

mrothauer:
Thanks to you for your support.

Issue #617, opened 2018-05-03
Safe TLD configurable


Moutard3:
The array containing the safe TLDs should be configurable with something like this:
`Configure::write('DebugKit', ['safeTLD' => ["localhost", "dev", "test", "example", "local","somethingelse"]]);`
I think that's a useful feature tu add one more layer to disable DebugKit in prod in case debug mode is not disabled, but it can't be used if not using a "standard" dev TLD (because we need to forceEnable it).
In my case I'm using `.loc` and Laragon is using `.oo` by default.

markstory:
If this config value is empty, would the code default to the current logic?

dereuromark:
It would be nice if that would disable it, solving the issue for people that dont like to use this overprotection. As it really can cause more harm than good for most setups.

dereuromark:
I explained that already in the other ticket.
It creates for those people (that are not experienced enough) the false assumption that all is good, but in fact debug mode is silently still on (without them knowing), and then every stack trace can reveal far more uncensored information about the underlying system.

markstory:
> In this case it is 10x more helpful to display it, and making sure that person knows that turning debug mode off is the right call (and fixes the root issue), not hiding the toolbar (fixing the symtom only).

I don't disagree with you here. The issue we were trying to solve is CakePHP (the project) being blamed/shamed when our users deploy applications in debug mode. It is *us* and not the individual developer who is blamed for these kinds of mistakes.

dereuromark:
Being a bit over-sensitive over a theoretical blame game is one thing, but actually compromising security of many apps even more this way due to the false security is a total different one.
For me this was a clear security issue opened - instead of closed. Because before it was a "baah, that was really obvious you fucked it up", now it is "CakePHP is dangerously hiding the toolbar by magic, that caused the accidental (unwilling) debug mode to expose lots of sensitive info over a long timeframe".

To sum things up:
For the first the user (beginner?) is to blame (not the framework clearly)
For the 2nd it gets now really hard to tell beginners that it was obvious, when its not (now the actual blame/shame is on the framework for making it that dangerous).

I stated those before the merge. And I was ignored :/

dereuromark:
To actually give a GOOD example of what should have been done:

On production (isSuspiciallyProduction true) the Icon should be even fatter, blinking, a huge arrow pointing to it with the bubble text of

> HELLO, you seem to be productive, but you still have debug on on"
For security reasons you MUST deactive it now.

[...]

PR #623, opened 2018-05-08
Make TLD configurable.

dereuromark:
Resolves https://github.com/cakephp/debug_kit/issues/617

I didnt move the defaults up as config defaults, so it currently merges instead of overwriting.
I am also fine with overwriting and empty config array to "skip" the check if people prefer to have that option rather than using  'safeTld' config to just "add". The current approach seems "safer", though.

### DIFF
--- a/src/ToolbarService.php
+++ b/src/ToolbarService.php
@@ -131,7 +131,8 @@ class ToolbarService
         }
 
         $tld = end($host);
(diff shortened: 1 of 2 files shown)


src/ToolbarService.php
```diff
@@ -64,6 +64,7 @@ class ToolbarService
             'DebugKit.Deprecations' => true,
         ],
         'forceEnable' => false,
+        'safeTld' => []
```
markstory: Any documentation or tests that could be updated for this new option?

Issue #631, opened 2018-06-10
Fatal error on CakePHP 3.6.5


asgraf:
After upgrading cakephp from 3.6.4 to 3.6.5 following fatal error started to appear:
> Fatal error: Cannot use Cake\Core\Plugin as Plugin because the name is already in use in [...]/vendor/cakephp/debug_kit/src/ToolbarService.php on line 16

This error does not occur on CakePHP 3.6.4 or earlier versions

How to reproduce:
1) Make a fresh installation of cakephp/app
2) Try to visit welcome page via browser or try to run bin/cake shell

saeideng:
confirm :+1: 

I could to see error via `bin/cake shell`  => with cake 3.6.5
but in web (browser ) not

saeideng:
related change https://github.com/cakephp/cakephp/commit/6b35611a

markstory:
Fixed in 3.16.3 😄

dereuromark:
This is the 2nd plugin. Will this now happen to everyone that adds this class?

markstory:
@dereuromark Only if they are also importing `Cake\Core\Plugin` in a top level class. Apparently I frequently do that and I don't write test around my plugin classes.

dereuromark:
But every app and plugin is doing that in the bootstrap, so it will effect every one, right?

markstory:
Its not a problem if the application imports Cake\Core\Plugin. Plugins will have this problem if they use root level namespaced classes like `AssetCompress\AssetScanner` but they will not have issues if they only use class names like `AssetCompress\Middleware\AssetCompressMiddleware` as the extra namespace prevents conflicts.

dereuromark:
In a rather fresh app, when now having

    Plugin::load('DebugKit', ['bootstrap' => true]);

with latest tag
you get

> Exception: Class 'DebugKit\ToolbarService' not found in [/.../demo.local/vendor/cakephp/debug_kit/config/bootstrap.php, line 26]

markstory:
@dereuromark I don't get that same error locally. The `DebugKit\ToolbarService` is still part of the plugin and has the correct class/filename.

dereuromark:
Interesting problem. This whole plugin loading of different plugins and different bootstrapping etc is a bit crazy.

saeideng:
>Fixed in 3.16.3

looks we can close it

PR #632, opened 2018-06-11
Fix duplicate class names caused by `DebugKit\Plugin` existing.

markstory:
Refs #631

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -11,7 +11,6 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 use Cake\Core\Configure;
-use Cake\Core\Plugin;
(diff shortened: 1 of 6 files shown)


config/bootstrap.php
```diff
@@ -64,10 +63,6 @@ function ($code, $message, $file, $line, $context = null) use (&$previousHandler
     ]);
 }
 
-if (Plugin::routes('DebugKit') === false) {
-    require __DIR__ . DS . 'routes.php';
-}
```
markstory: The Plugin class will take care of this now.

saeideng: this required and reverted here  
https://github.com/cakephp/debug_kit/commit/aabaecb032f7e91

markstory: Yeah I goofed up.

PR #633, opened 2018-06-15
Accept Cake\Database\Query for sql()/sqld()

garas:
sql()/sqld() should accept both `Cake\ORM\Query` (`TableRegistry::get('Articles')->find()`) and `Cake\Database\Query` (`ConnectionManager::get('default')->newQuery()`)


lorenzo:
Thanks!

Issue #624, opened 2018-05-14
Cake 3.6.* & ES Plugin 1.* causes a type error


burzum:
> Argument 1 passed to Elastica\Client::setLogger() must implement interface Psr\Log\LoggerInterface, instance of DebugKit\Database\Log\DebugLog given, called in C:\xampp\htdocs\wa3\vendor\cakephp\debug_kit\src\Panel\SqlLogPanel.php on line 67

This happens with debug kit 3.16.0 but not with 3.14.2. And that's all I've changed to make it work again. So there must be something wrong with 3.16.0. Guess a type was added somewhere.

makamo:
Same with UseMuffin/Webservice

markstory:
It is likely this line https://github.com/cakephp/debug_kit/blob/e7d7fc3bce3a89b5fc57d34ca8666175c52e3d4a/src/Panel/SqlLogPanel.php#L67

which would have been changed to accommodate the deprecated `logger()` method on `Connection`. This method name is clearly in use by other classes.

burzum:
@markstory this explains the problem, but doesn't provide a solution?

davidyell:
How come this has been closed without a resolution?

josegonzalez:
Fixed on the other end? https://github.com/cakephp/elastic-search/pull/156

ADmad:
@josegonzalez The issue affects other community code too, for e.g. mufin/webservice. IMO the fix should be done in debugkit.

lorenzo:
Didn’t the pull request I merged fix his?

davidyell:
@lorenzo If it did, then it needs a release. I've just created a new fresh application on `3.6.2` with the Webservice plugin, and when I force enable DebugKit, I get a fatal for the logger.

I would recommend this issue be re-opened.

ceeram:
wrt muffin-webservice plugin, we can probably get by this by adding this in its Connection class:

```
    /**
     * no-op
     *
     * to avoid non-PSR loggers (Crud, DebugKit) getting passed to webservice driver, which requires PSR implementations
     */
    public function setLogger()
    {

    }
```

davidyell:
> This method name is clearly in use by other classes outside of CakePHP.

Yes, by the PSR3 Logger at least.

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md#4-psrlogloggerawareinterface

Nic0tiN:
Hello,

[...]

PR #636, opened 2018-08-03
Fix issue #624

Nic0tiN:
As per discussion in thread #624

### DIFF
--- a/src/Panel/SqlLogPanel.php
+++ b/src/Panel/SqlLogPanel.php
@@ -64,7 +64,9 @@ class SqlLogPanel extends DebugPanel
             $logger = new DebugLog($logger, $name, $includeSchemaReflection);
 
             $connection->logQueries(true);
-            $connection->setLogger($logger);
+            if (method_exists($connection, 'setLogger')) {
+                $connection->setLogger($logger);
+            }
             $this->_loggers[] = $logger;
         }
     }


dereuromark:
Doesn't that silently cause problems for classes that follow the psr and do not offer this method?

Nic0tiN:
I haven't been gone through the whole code but I don't see where it could be a problem.

However, a type hinting exception will be raised when a class is PSR compliant and offers this method.
If we take UseMuffin\Webservice as an example, it doesn't because we use `method_exists` on `Connection::__call` that proxies to PSR compliant class `AbstractDriver`. That's excatly the problem we were trying to solve but may be that solution is too exclusive to UseMuffin\Webservice...

dereuromark:
This is for sure better than before :) I would say.

src/Panel/SqlLogPanel.php
```diff
@@ -64,7 +64,9 @@ public function initialize()
             $logger = new DebugLog($logger, $name, $includeSchemaReflection);
 
             $connection->logQueries(true);
-            $connection->setLogger($logger);
+            if (method_exists($connection, 'setLogger')) {
+                $connection->setLogger($logger);
+            }
```
ADmad: You are missing the "else" part which should call the old method `$connection->logger($logger);`

ADmad:
@Nic0tiN Have you tried this patch and checked that it fixes the problem?

Nic0tiN:
@ADmad Yes, I use those new files within my project base on Cakephp 3.6.7, DebugKit 3.16.3, Webservice (dev-master).

davidyell:
I'd like to see a test for this to prove compatibility with PSR3

PR #646, opened 2018-10-13
prepaire to 4.x 

saeideng:


config/bootstrap.php
```diff
@@ -69,17 +68,11 @@ function ($code, $message, $file, $line, $context = null) use (&$previousHandler
...
-    });
-} else {
-    // Setup dispatch filter
-    $debugBar = new DebugBarFilter(EventManager::instance(), (array)Configure::read('DebugKit'));
-    $debugBar->setup();
-    DispatcherFactory::add($debugBar);
-}
+EventManager::instance()->on('Server.buildMiddleware', function ($event, $queue) use ($service) {
+    $middleware = new DebugKitMiddleware($service);
+    $queue->insertAt(0, $middleware);
```
markstory: We could move this to the Plugin class' `middleware` hook now.

src/Plugin.php
```diff
@@ -14,10 +14,82 @@
...
+        $service = new ToolbarService(EventManager::instance(), (array)Configure::read('DebugKit'));
+        $this->service = $service;
+
+        if (!$service->isEnabled() || php_sapi_name() === 'cli' || php_sapi_name() === 'phpdbg') {
+            return;
+        }
+
+        $this->setDeprecationHandler($service);
+
+        if (!$this->isEnabled('routes')) {
```
ADmad: This is not necessary, there's already a routes() hook method which will handle routes file inclusion.

lorenzo: Remove it then?

ADmad: Yes

ADmad:
`templates/layout/Email/text/default.php` needs to be reanmed to `templates/layout/email/text/default.php`.

saeideng:
>templates/layout/Email/text/default.php needs to be renamed to templates/layout/email/text/default.php.

if you meant in test_app , I did.

, i will open another PR for replacing `Shell` with `Command`

PR #648, opened 2018-11-09
- Error in Plugin and Controller name

yasnaderi:


Lib/Panel/HistoryPanel.php
```diff
@@ -67,8 +67,8 @@ public function beforeRender(Controller $controller) {
 				$historyStates[] = array(
 					'title' => $title,
 					'url' => array_merge($prefix, array(
-						'plugin' => 'debug_kit',
-						'controller' => 'toolbar_access',
+						'plugin' => 'DebugKit',
+						'controller' => 'ToolbarAccess',
```
markstory: Underscore values generally work find in 2.x why does this need to change?

yasnaderi: In SQL log when click on the button, does not work for me.

image

image

PR #653, opened 2018-12-01
Make DebugKitTransport a better proxy

### LINKED ISSUE #558: DebugKitTransport and custom methods on TransportClasses
We have some custom mail transports, with some custom methods. When running with debug_kit and calling them in controllers, it errors out with unknown method since DebugKitTransport is injected instead of the normal Transport and returned with `$email->getTransport()`

Is there a better way to avoid this than calling 

`Configure::write('DebugKit.panels', ['DebugKit.Mail' => false]);`

markstory:
Add magic method support to DebugKitTransport so it can proxy transports that have additional properties and methods.

Fixes #558


saeideng:
@markstory  this PR is only for debug_kit 3.17 ? (cake 3.7 )
or could debug_kit 3.16.x (cake3.6) have this ?

markstory:
@saeideng It could, I was being lazy and didn't back port it.

markstory:
Looks good to me.

PR #664, opened 2019-03-29
Added turbolinks compatibility

### LINKED ISSUE #663: Debugkit does not work with Turbolinks
Turbolinks are an improvement in any webapplication.
We use Turbolinks in a current project and recognized, that debugkit is not compatible with Turbolinks due its usage of DOMContentLoaded.

To prevent this issue, i recommend to just add another eventlistener on "turbolinks:load" to make it compatible. It should also check if the "turbolinks:load" listeners are already added to prevent a doubled execution of the handler.

What is Turbolinks:
> Turbolinks® makes navigating your web application faster. Get the performance benefits of a single-page application without the added complexity of a client-side JavaScript framework. Use HTML to render your views on the server side and link to pages as usual. When you follow a link, Turbolinks automatically fetches the page, swaps in its <body>, and merges its <head>, all without incurring the cost of a full page load.
_[Turbolinks](https://github.com/turbolinks/turbolinks)_

D4rkMindz:
Fixes #663
- Added listeners
- Added check if llisteners already applied

### DIFF
--- a/webroot/js/toolbar.js
+++ b/webroot/js/toolbar.js
@@ -97,6 +97,13 @@ if (elem) {
     doc.addEventListener('DOMContentLoaded', onReady, false);
     doc.addEventListener('DOMContentLoaded', proxyAjaxOpen, false);
     doc.addEventListener('DOMContentLoaded', proxyAjaxSend, false);
+
+    if (!window.__debugKitTurbolinksListenerApplied) {
+      doc.addEventListener('turbolinks:load', onReady, false);
+      doc.addEventListener('turbolinks:load', proxyAjaxOpen, false);
+      doc.addEventListener('turbolinks:load', proxyAjaxSend, false);
+      window.__debugKitTurbolinksListenerApplied = true;
+    }
   } else {
     throw new Error('Unable to add event listener for DebugKit. Please use a browser' +
       'that supports addEventListener().');
--- a/webroot/js/toolbar.js
+++ b/webroot/js/toolbar.js
@@ -92,13 +92,20 @@ if (elem) {
     };
   };
 
-  // Bind on ready callback.
+  // Bind on ready callbacks to DOMContentLoaded (native js) and turbolinks:load
+  // Turbolinks replaces the body and merges the head of an HTML page.
+  // Since the body is already loaded (DOMContentLoaded), the event is not triggered.
   if (doc.addEventListener) {
+    // This ensures that all event listeners get applied only once.
     if (!window.__debugKitListenersApplied) {
+      // The DOMContentLoaded is for all pages that do not have Turbolinks
       doc.addEventListener('DOMContentLoaded', onReady, false);
       doc.addEventListener('DOMContentLoaded', proxyAjaxOpen, false);
       doc.addEventListener('DOMContentLoaded', proxyAjaxSend, false);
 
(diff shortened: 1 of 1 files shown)


webroot/js/toolbar.js
```diff
@@ -97,6 +97,13 @@ if (elem) {
     doc.addEventListener('DOMContentLoaded', onReady, false);
     doc.addEventListener('DOMContentLoaded', proxyAjaxOpen, false);
     doc.addEventListener('DOMContentLoaded', proxyAjaxSend, false);
+
+    if (!window.__debugKitTurbolinksListenerApplied) {
```
D4rkMindz: This checks if the debugkit listeners are already applied

webroot/js/toolbar.js
```diff
@@ -97,6 +97,13 @@ if (elem) {
...
+      doc.addEventListener('turbolinks:load', onReady, false);
+      doc.addEventListener('turbolinks:load', proxyAjaxOpen, false);
+      doc.addEventListener('turbolinks:load', proxyAjaxSend, false);
```
D4rkMindz: These are the required listeners

webroot/js/toolbar.js
```diff
@@ -97,6 +97,13 @@ if (elem) {
...
+      window.__debugKitTurbolinksListenerApplied = true;
```
D4rkMindz: Set the check to true to not apply the listeners again

webroot/js/toolbar.js
markstory: When would this block be re-run?

D4rkMindz: Literally on every page navigation. The result of that is that the same listener (e.g. `onReady`) doesn't get added multiple times. Otherwise the listener will be added multiple times and the e.g. `onReady`-method gets called <page navigations done>^2 times which can cause serious bugs.

The check could also wrap the `DOMContentLoaded` listeners, but its not required. It would "just" improve performance.

webroot/js/toolbar.js
```diff
@@ -94,15 +94,15 @@ if (elem) {
 
   // Bind on ready callback.
   if (doc.addEventListener) {
-    doc.addEventListener('DOMContentLoaded', onReady, false);
-    doc.addEventListener('DOMContentLoaded', proxyAjaxOpen, false);
-    doc.addEventListener('DOMContentLoaded', proxyAjaxSend, false);
+    if (!window.__debugKitListenersApplied) {
```
D4rkMindz: I decided to also wrap the `DOMContentLoaded` listeners to improve performance

For further information see [my previous comment](https://github.com/cakephp/debug_kit/pull/664/commits/c05cfbf23c128c152425d100b7cad0c08964d28a#r270721674)

webroot/js/toolbar.js
ravage84: I guess they are requied by turbolinks. Could you please add a link in a comment on top of these, please?

webroot/js/toolbar.js
```diff
@@ -94,9 +94,16 @@ if (elem) {
 
   // Bind on ready callback.
```
ravage84: I think we should update the comment here. The code has now a much higher implicit complexity than before.

Also the current comment only documents what we do,not why.
@D4rkMindz could you please rephrase it with the necessary details/background info?

PR #667, opened 2019-04-06
Fix email rendering.

markstory:
We were using recently broken APIs. Update to the new Mailer/Message interface

### DIFF
--- a/src/Mailer/PreviewResult.php
+++ b/src/Mailer/PreviewResult.php
@@ -49,7 +49,7 @@ class PreviewResult extends AbstractResult
         $mailer->render();
         $message = $mailer->getMessage();
         $this->parts = [
-            'text' => $message->getBody(Message::MESSAGE_HTML),
+            'html' => $message->getBody(Message::MESSAGE_HTML),
             'text' => $message->getBody(Message::MESSAGE_TEXT),
         ];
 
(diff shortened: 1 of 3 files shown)


src/Plugin.php
```diff
@@ -61,7 +61,7 @@ public function bootstrap(PluginApplicationInterface $app): void
     public function middleware(MiddlewareQueue $middleware): MiddlewareQueue
     {
         if ($this->service) {
-            $middleware->add(new DebugKitMiddleware($this->service));
+            $middleware->insertAt(0, new DebugKitMiddleware($this->service));
```
markstory: This lets the toolbar display again :+1:

src/Mailer/PreviewResult.php
```diff
@@ -47,14 +46,14 @@ protected function processMailer(Mailer $mailer, $method)
...
-        $reflection = new ReflectionClass($mailer);
-        $prop = $reflection->getProperty('_email');
-        $prop->setAccessible(true);
-        $email = $prop->getValue($mailer);
-
-        $this->parts = $email->getRenderer()->renderTemplates($email, '');
+        $mailer->render();
+        $message = $mailer->getMessage();
+        $this->parts = [
+            'text' => $message->getBody(Message::MESSAGE_HTML),
```
garas:
```suggestion
            'html' => $message->getBody(Message::MESSAGE_HTML),
```

PR #668, opened 2019-04-09
Exclude more files from being exported

ravage84:
I was toying around with a deployment script and noticed these unnecessary (in a deployment context) files.


garas:
There is no `CONTRIBUTING.md` and `Makefile`.

List could be sorted/grouped as it is in almost random order.

Issue #670, opened 2019-04-15
Use default custom RouteClass of plugin broke the DebugKit


Aerue:
Working with a custom Router into a plugin result in 404 errors with the debug kit plugin.

**config/routes.php**

    Router::defaultRouteClass(\Plugin\Routing\Route\MyCustomRouter::class);

Fix :

**debug_kit/config/routes.php**

    ...
    $routes->setExtensions('json');
    $routes->setRouteClass(\Cake\Routing\Route\DashedRoute::class);
    $routes->connect(
        '/toolbar/clear-cache',
        ['controller' => 'Toolbar', 'action' => 'clearCache']
    );
    ...

markstory:
Any chance you'd be willing to send a pull request for this?

ADmad:
Closing as related PR is open.

PR #672, opened 2019-04-16
Fix issue #670

Aerue:


ADmad:
Fix looks good, just the CS error needs to be fixed.

markstory:
Thank you! I fixed the cs error before merging.

PR #673, opened 2019-04-27
Move documentation into sphinx

markstory:
Use the new plugin docs build tooling for DebugKit docs.


garas:
Those few GIFs increases repo size too much.

Files 640KB -> 50 MB, files can be removed in future, but `.git` directory increase is permanent.

`.git` directory 2MB -> 14MB without shrinking possibility in future.

I would consider remove (with forced git push) those GIFs before merging to `master`.

markstory:
How do you propose the gifs be included in the documentation? We could link to the images in the cakephp/docs repo but it could be fragile in the future.

garas:
First I'm considering if they add enough value to content.

To save file space, they could be in shared `_static` directory, because currently they are exact copies, takes 12.5 x 4 = 50 MB, one copy already would be much better.

Also, they could be converted to `mp4` (mailer gif reduced from 3.36MiB to 573.38KiB mp4 (-83.35%) [here](https://ezgif.com/)), and embedded as HTML5 video. This would also let viewer to free control of video playback, gifs have no playback control. There is Sphinx plugin for [html5 videos](https://github.com/sphinx-contrib/video).

With both optimizations, they would take only few MBs and would be acceptable size increase.

markstory:
The mp4 plugin could work well. I will give that a try. I can also de-duplicate the files and copy them during the build.

markstory:
@garas There is now only one copy of the mp4 files in the repo and all languages reference the same video files.

garas:
It should be good now.

But videos are broken, they play second or two.

I've compared with what I get from that online converter, they are missing few bytes, `\r\n` was converted to `\n`.

Add `mp4` as binary to `.gitattributes`.

garas:
Re-add videos to git, they are still broken.

markstory:
@garas I'll re-generate them tonight.

PR #674, opened 2019-05-04
Upgrade to phpunit8

markstory:


tests/TestCase/Panel/SqlLogPanelTest.php
```diff
@@ -88,7 +89,10 @@ public function testInitializePassesIncludeSchema()
         $db = ConnectionManager::get('test');
         $logger = $db->getLogger();
         $this->assertInstanceOf('DebugKit\Database\Log\DebugLog', $logger);
-        $this->assertAttributeEquals(true, '_includeSchema', $logger);
```
ADmad: The core has lot of such assertions for protected properties which use deprecated assert methods like `assertAttributeEquals()`.

We could include this [AccessibilityHelperTrait](https://github.com/FriendsOfCake/cakephp-test-utilities/blob/master/src/AccessibilityHelperTrait.php) in core itself to help easily update such tests.

PR #675, opened 2019-05-06
fix return type in console

saeideng:


markstory:
I think I got all the failing tests fixed.

PR #678, opened 2019-05-24
Fix query DebugLog

cleptric:
Make debug_kit work again after cakephp/cakephp#13276
Not sure about the dock block. Can't find where it gets inherit from.


ADmad:
I updated `DebugLog` as per core changes.

ADmad:
The test causing failure for PR build has been fixed on `4.x`.

PR #681, opened 2019-05-30
Update middleware to implement MiddlewareInterface.

ADmad:


ADmad:
Thanks for taking care of the view option changes.

PR #689, opened 2019-07-06
Fix string to bool conversion

Fuxy22:
Countary to any sane person's expectations in php (bool)"false"
does not evaluate to "false" but "true" instead

### DIFF
--- a/src/Controller/ComposerController.php
+++ b/src/Controller/ComposerController.php
@@ -65,7 +65,7 @@ class ComposerController extends Controller
         $input = new ArrayInput([
             'command' => 'outdated',
             '--no-interaction' => true,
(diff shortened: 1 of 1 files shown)


ravage84:
Proof: https://3v4l.org/b3p4n

ravage84:
I see we do not have any tests for this controller:

https://github.com/cakephp/debug_kit/tree/e2d856ed7c504b80354038b22d6f342c13ac5145/tests/TestCase/Controller

Might be a good idea to add a regression test.
What do you think, @Fuxy22 ?

Fuxy22:
Testing is good although I'm not very familiar with how its done in cakephp

markstory:
Merging. I'll circle back and add tests for the entire controller later this week.

PR #693, opened 2019-08-10
Hide debugkit routes with a toggle

markstory:
Alternate solution to #553 that hides the routes but includes a toggle to show them if you want to see them. I've also updated the summary number as it shouldn't include debugkit routes which are hidden by default.

I've also made buttons a bit bigger as they were really small before.

### DIFF
--- a/src/Panel/RoutesPanel.php
+++ b/src/Panel/RoutesPanel.php
@@ -37,7 +37,7 @@ class RoutesPanel extends DebugPanel
         }
 
         $routes = array_filter(Router::routes(), function ($route) {
-            return $route->defaults['plugin'] === 'DebugKit';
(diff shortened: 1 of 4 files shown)


src/Panel/RoutesPanel.php
```diff
@@ -36,7 +36,11 @@ public function summary()
             return 0;
         }
 
-        return count(Router::routes());
+        $routes = array_filter(Router::routes(), function ($route) {
+            return strpos($route->getName(), 'debugkit.') !== 0;
```
garas: Matching on `$route->defaults['plugin']` would avoid string operations.

garas:
Travis fails because `travis.yml` needs `services:` option

PR #695, opened 2019-08-17
Fix deprecation for isNew(bool)

Moutard3:
In cake 4.0 isNew as setter is deprecated, setNew should be used.


saeideng:
thank you

garas:
The line should be removed at all. Entity created by `newEntity()` is already new.

PR #698, opened 2019-09-08
Fix unknown Message::message method

Moutard3:
The method got removed in https://github.com/cakephp/cakephp/pull/13536/commits/aeac7a45b675810ea10bb419f402d89bdf938ece


markstory:
Thanks :+1:

PR #699, opened 2019-09-10
Update App::path() usages

ADmad:
Refs cakephp/cakephp#13618


src/Controller/MailPreviewController.php
```diff
@@ -183,7 +184,7 @@ protected function getMailPreviewClasses()
                 return $plugin === 'DebugKit';
             })
             ->map(function ($plugin) {
-                return [App::path('Mailer/Preview', $plugin), "$plugin."];
+                return [[Plugin::classPath($plugin) . 'Mailer/Preview/'], "$plugin."];
```
markstory: This should be `CorePlugin` and then the extra import on line22 can be removed.

PR #702, opened 2019-10-08
run VACUUM command to free disc space

davalb:
I noticed that the size of `debug_kit.sqlite` will keep growing even though the `RequestsTable` has a garbage collection implemented. 

After some research (https://stackoverflow.com/questions/35787220/sqlite-file-size-not-decreasing-after-deleting-some-rows-from-database-in-ios and https://sqlite.org/lang_vacuum.html) I found out that disc space is not automatically freed. It is only freed if the database is recreated with the VACUUM; command.

### DIFF
--- a/src/Model/Table/RequestsTable.php
+++ b/src/Model/Table/RequestsTable.php
@@ -13,6 +13,7 @@
 namespace DebugKit\Model\Table;
 
 use Cake\Core\Configure;
+use Cake\Database\Driver\Sqlite;
 use Cake\ORM\Query;
 use Cake\ORM\Table;
 use DebugKit\Model\Entity\Request;
@@ -108,5 +109,10 @@ class RequestsTable extends Table
 
         $statement = $query->execute();
         $statement->closeCursor();
+
+        $conn = $this->getConnection();
+        if ($conn->getDriver() instanceof Sqlite) {
+            $conn->execute('VACUUM;');
+        }
     }
 }


src/Model/Table/RequestsTable.php
```diff
@@ -108,5 +109,7 @@ public function gc()
 
         $statement = $query->execute();
         $statement->closeCursor();
+        $conn = ConnectionManager::get($this->defaultConnectionName());
```
garas:
```suggestion
        $conn = $this->getConnection();
```

davalb: Thank you, I made the change.

davalb:
Wait a sec, I just had the thought, what if somebody doesn't use sqllite? the VACUUM command might lead to an error in other database-engines. I believe I need to check what database-engine is used, before runnign VACUUM. Sorry for the trouble

src/Model/Table/RequestsTable.php
```diff
@@ -108,5 +108,10 @@ public function gc()
 
         $statement = $query->execute();
         $statement->closeCursor();
+
+        $conn = $this->getConnection();
+        if (is_a($conn->getDriver(), 'Cake\Database\Driver\Sqlite')) {
```
markstory: We generally use `instanceof` instead of `is_a`

davalb: ok, I changed that!

PR #704, opened 2019-10-30
Use sub query instead of ids array.

ADmad:


markusramsak:
doesn't work in MySQL because LIMIT is not supported in a subquery in MySQL.

PR #705, opened 2019-10-30
Right align toolbar tabs.

ADmad:
Since the toolbar icon is placed on the right, having panels left aligned means one has to unnecessarily move the mouse towards left of screen, on highish resolutions with few panels, even to click the rightmost panel, after clicking the icon.


ADmad:
Before:
Screenshot_2019-10-30 Header Slides - CMS Cake4

After:
Screenshot_2019-10-30 Header Slides - CMS Cake4(1)

markstory:
I'll figure out why the tests are failing.

Issue #578, opened 2018-01-11
invalid request history 


saeideng:
sometimes you see extra request in the debug_kit that is invalid

![debug_kit_asset](https://user-images.githubusercontent.com/8215801/34848154-fb2083ac-f732-11e7-8a15-eae7a529af89.png)

also number of request history  limited to `10`
with this invalid request you lost previous request history

suggestions:
1) adding filters buttons to filter (/remove )  assets (js/css,...) 
2) adding a  button to remove invalid histories ( runtime ) , and reload list
3) capability to increase limitation from `10` 
4) pagination ( @markstory )

I like 2

markstory:
Are these requests handled by the asset middleware?  We could add pagination to the request history as another option.

saeideng:
>Are these requests handled by the asset middleware?

```php
//APP/Application.php
    public function middleware($middlewareQueue)
    {
        $middlewareQueue
            // Catch any exceptions in the lower layers,
            // and make an error page/response
            ->add(ErrorHandlerMiddleware::class)

            // Handle plugin/theme assets like CakePHP normally does.
            ->add(AssetMiddleware::class)

            // Add routing middleware.
            ->add(new RoutingMiddleware($this));

        return $middlewareQueue;
    }
```

this problem occur in several my apps (probably in all )
with cake 3.5, 3.4 ,....

markstory:
Ok. Are those assets coming from plugins? That would explain why cake is handling them and your webserver is not serving them directly.

saeideng:
>Ok. Are those assets coming from plugins? 

Yep, local plugin 
I use `plugins/... ` for templates 

for example `plugins/BackEnd/webroot/bootstrap/css/bootstrap.min.css` for above screenshot

saeideng:
@markstory  I tested 
I added `touch(time().'_.txt');` into first of  `webroot/index.php`
and I try load 2 css directly from web

I add new/fresh plugin for test 
```
/config/bootstrap.php
Plugin::load('New'); // this plugin just have 1 css file 

//browser
http://site.local/test/new/css/test.css // creates `1515785147_.txt`
http://site.local/test/css/test.css // OK (without creating `txt` file )

markstory:
Ok. It makes sense that debugkit would be logging asset requests as they are being handled by CakePHP. Would being able to paginate through history help solve your problem?

saeideng:
also I added 
`touch(time().'_asset');` into `AssetMiddleware::__invoke()`
[...]

PR #707, opened 2019-11-02
don't save plugins assets

saeideng:
a solution for #578

### DIFF
--- a/src/ToolbarService.php
+++ b/src/ToolbarService.php
@@ -21,6 +21,7 @@ use Cake\Http\ServerRequest;
 use Cake\Log\Log;
 use Cake\ORM\TableRegistry;
 use Cake\Routing\Router;
+use Cake\Utility\Inflector;
 use DebugKit\Panel\PanelRegistry;
 use Psr\Http\Message\ResponseInterface;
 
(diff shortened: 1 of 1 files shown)


saeideng:
another option could be add a button to hide this and +  pagination requests

src/ToolbarService.php
```diff
@@ -214,6 +215,15 @@ public function saveData(ServerRequest $request, ResponseInterface $response)
         ) {
             return null;
         }
+        // Skip plugins assets
+        if (strpos($path, '.') !== false &&
+            strpos($path, '/.') === false &&
+            strpos($path, '..') === false &&
+            $response->getStatusCode() === 200 &&
+            $this->checkFileExists($path)
```
markstory: Instead looking for the file, could we skip all css, js and common image files? The checks would be simpler.

saeideng: first I wanted to not catch all files that not served by AssetMiddleware
https://github.com/cakephp/cakephp/blob/61a1ff9672430456a7a261f053911cab93793742/src/Routing/Middleware/AssetMiddleware.php#L86-L93
and then people should see other requests that are 404 not exists
just like the main assets behaviors in DebugKit 

but if `$response->getStatusCode() === 200 &&` is enough probably we could do this method simpler

dereuromark:
Do we want to proceed here? Incremental Improvements?

saeideng:
closing in favor of #764

PR #712, opened 2019-11-10
Fix errors reported by pslam

ADmad:
The first time I ran pslam it threw 123 errors :sob:.  There are probably still few valid errors which need to be fixed and ignores removed from baseline but that can be done in due time.


dereuromark:
> Cake\Filesystem\File is marked deprecated

We really shouldn't mark those as deprecated in 3.x or 4.x
It was a bad idea from the start, given that there is currently not even remotely a replacement for those IMO.

src/View/Helper/TidyHelper.php
```diff
@@ -138,25 +137,26 @@ public function tidyErrors($in = '', &$out = '')
...
-        $path = $File->pwd();
-        $errors = $path . '.err';
-        $this->_exec("tidy -eq -utf8 -f $errors $path");
-        $File->delete();
-
-        if (!file_exists($errors)) {
+        $filePath = rtrim(TMP, DS) . DS . uniqid() . '.html';
+        file_put_contents($filePath, $out);
+        $errorFilePath = $filePath . '.err';
+        $this->_exec("tidy -eq -utf8 -f $errorFilePath $filePath");
```
garas: `TMP` could contain spaces, so arguments should be escaped.

But anyway, I only find reference to this class in `src/View/AjaxView.php` via `@property`, but no further usage. This class can be deleted.

garas: Added f6a4479 2012-04-14 Tidy Panel
Removed 5c20afa 2012-05-12 Remove tidy panel

ADmad: Yeah it did seem to me that the class was unused but I kept my focus to just fixing errors reported by psalm.

I can remove it if @markstory confirms.

markstory: I think this was intended as an optional panel. I personally don't use it, perhaps @dereuromark can speak to how useful he finds it?

dereuromark: I must say I didnt use it much at all :)

markstory: We could remove it and see if anyone notices :smile:

ADmad: Done

PR #713, opened 2019-11-22
Fix CS

dereuromark:


dereuromark:
A future PR can take care of the new missing ones once stickler is not sabotaging it.

Issue #718, opened 2019-12-18
Wrong code/context display when the deprecations panel is on


berarma:
I'm using DebugKit 3.21.0 on CakePHP 3.8.7.

When the deprecations panel is on and an application error is displayed, the code and context is not from the app code that caused the error but from a function in DebugKit's bootstrap file.

```
Code Context
            if ($previousHandler) {
                return $previousHandler($code, $message, $file, $line, $context);
            }

Cake\Core\BasePlugin::{closure}() - ROOT/vendor/cakephp/debug_kit/config/bootstrap.php, line 42
```

The deprecation panel is on by default. I think this is happening since the deprecation panel was added maybe.

markstory:
Where are you seeing those deprecation warnings. DebugKit should be capturing deprecations and displaying them in the deprecation panel, and from a quick check locally messages in that panel look correct.

How can I reproduce the incorrect results you're seeing?

othercorey:
Is this related to https://github.com/cakephp/cakephp/issues/14068?

markstory:
@othercorey  I don't think so, but it is a similar class of problem caused by the stack traces being wrong on errors.

othercorey:
> @othercorey I don't think so, but it is a similar class of problem caused by the stack traces being wrong on errors.

FYI, I pulled that issue from a screenshot using debug kit, I believe. I have't had time to setup the failure to check.

PR #722, opened 2019-12-29
Add trace_offset to error handler

markstory:
Using the trace_offset will allow wrapped errors to have the correct
code and context displayed.

Refs #718
Refs cakephp/cakephp#14090


othercorey:
Seems ok to me.  These things are always a little hacky because they are ... hacking it.

PR #732, opened 2020-01-27
Update index.rst

davidspeijer:
For new Cakephp users it is not clear that database config for debug_kit should be placed under Datasources.

Should the file reference "config/app.php" also be changed to config/app_local.php ?


markstory:
Thank you! I don't think mentioning app_local.php here is necessary.

Issue #729, opened 2020-01-17
Authorization fail


Erwane:
When using Cakephp/Authorization plugin, debugkit fail.
One solution is to skip authorization in AuthorizationMiddleware (https://github.com/cakephp/authorization/issues/80#issuecomment-575626974) , but it's not a good place to add exceptions.

One way is to check this in DebugKit/Controller but they don't extends a common "AppController" (in plugin) so, there is a lot of "replicated code". Below, an example in the "beforeFilter" method

```php
// In DebugKit.Controller::beforeFilter()
$authorizationService = $this->getRequest()->getAttribute('authorization');
if ($authorizationService instanceof \Authorization\AuthorizationService) {
    $service = $authorizationService->skipAuthorization();
}
```

Do you think we can add this common plugin controller with no risk ?

markstory:
We could, I am not sure we should though. Having debugkit disable authentication and authorization checks is risky as people often forget to disable debugkit in production. Perhaps we could do this via an opt-in option though?

dereuromark:
@markstory You could add if (Configure::read(debug)) into it?

Erwane:
An opt-in options is like configuring the `DebugKit.safeTld` or `DebugKit.forceEnable` i think.

But i'm not sure it's a risk. The Authorization plugins is only disable for the DebugKit request and it's a XHR request. You can't access app url by a DebugKit call.

markstory:
That could work, I always worry about the person who accidentally leaves debug on in production. As it is cake not the user that gets blamed.

Erwane:
As i said, i'm sure it's not a problem. Disabling Authorization plugin in Debugkit controllers only disable it for Debukit requests, not app requests

Erwane:
i try a PR ?

markstory:
@Erwane That would be great!

othercorey:
I suppose the question is do we expose something that people can use to probe websites with to gain environment data/etc.

If a user has DebugKit enabled, what level of guarantee do we provide?

LordSimal:
We documented the need to set `DebugKit.ignoreAuthorization` to true in the README.md of this repo as well as in the docs for the debug_kit.

Therefore I will close this issue.

PR #734, opened 2020-02-03
#729 v3 authorization fail

Erwane:
Skip cakephp/authorization plugin for DebugKit requests only


tests/TestCase/Controller/DebugKitControllerTest.php
```diff
@@ -0,0 +1,79 @@
...
+        Configure::write('debug', false);
+
+        $request = new ServerRequest(['url' => '/debug-kit/']);
+        $controller = new DebugKitController($request, new Response());
+        $event = new Event('testing');
+
+        // try/catch/finally instead of expectExcetion
+        // to restore `debug` configuration
+        try {
+            $controller->beforeFilter($event);
```
markstory: Integration tests generally make requests to the application and check the response contents.

tests/TestCase/Controller/DebugKitControllerTest.php
```diff
@@ -0,0 +1,79 @@
...
+        } catch (Exception $e) {
+            $this->assertInstanceOf('Cake\Http\Exception\NotFoundException', $e);
+            $this->assertSame('Not available without debug mode on.', $e->getMessage());
+        } finally {
+            Configure::write('debug', $oldStatus);
```
markstory: You don't need to do this. The test harness will reset `Configure` state during its teardown method.

src/Controller/DebugKitController.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function beforeFilter(Event $event)
+    {
+        // TODO add config override.
+        if (!Configure::read('debug')) {
+            throw new NotFoundException('Not available without debug mode on.');
+        }
+
+        // Skip authorization for DebuKit requests
+        $authorizationService = $this->getRequest()->getAttribute('authorization');
```
markstory: I thought this was going to be a behavior that users could turn on. Something like `DebugKit.safeTLD` works.

Erwane:
Thanks. I thought my code was clean ;)
I will do better.

tests/TestCase/Controller/DebugKitControllerTest.php
Erwane: I'm ok with you about `testDebugDisabled` but for the `testSkipAuthorization` i should test the AuthorizationService::authorizationChecked() status, and i only can do that thrue controller request.
Any ideas ?

src/Controller/DebugKitController.php
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        // If CakePHP Authorization\Authorization plugin is enabled,
+        // ignore it, only if `DebugKit.ignoreAuthorization` is set to true
+        $authorizationService = $this->getRequest()->getAttribute('authorization');
+        if ($authorizationService instanceof \Authorization\AuthorizationService) {
+            if (Configure::read('DebugKit.ignoreAuthorization')) {
+                $authorizationService->skipAuthorization();
+            } else {
+                Log::info(
```
markstory: :+1: This is a nice inclusion.

Issue #747, opened 2020-04-10
Remove cakephp/plugin-installer dependency


othercorey:
Can we remove this dependency? Does any code in debug_kit need it?

ADmad:
It doesn't, unless debugkit is doing something special. 

The plugin installer readme states:
> Your plugins themselves do not need to require cakephp/plugin-installer

dereuromark:
The app has it
https://github.com/cakephp/app/blob/master/composer.json#L11

Technically, it is a bit cleaner to have the dependency on the plugins, thus auto requiring it from the app perspective.
But I can see that this is a bit annoying to do for all plugins.
So app only seems fine.

PR #748, opened 2020-04-10
Remove unneeded dependency

dereuromark:
Resolves https://github.com/cakephp/debug_kit/issues/747

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,6 @@
         "php": ">=7.2",
         "cakephp/cakephp": "^4.0",
         "cakephp/chronos": "^2.0",
-        "cakephp/plugin-installer": "^1.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -26,7 +26,6 @@
         "php": ">=7.2",
         "cakephp/cakephp": "^4.0",
         "cakephp/chronos": "^2.0",
-        "cakephp/plugin-installer": "^1.0",
         "composer/composer": "^1.3",
```
othercorey, line 29: I wonder if the composer dependency was added just for plugin-installer.

dereuromark: No, several classes are using composer classes

    use Composer\Json\JsonFile;

etc

othercorey: Makes sense!

Issue #746, opened 2020-04-10
Inconsistent names for timer entries


othercorey:
When looking at a profile, you see inconsistent prefix formats:

- `Event:`
- `View Render`
- `Controller action`
- `Twig:`

Can we define a consistent way to grouping category > sub-category names? Should we just use `:` to separate them?

markstory:
Using : for key namespacing makes sense to me.

PR #750, opened 2020-04-12
Updated timer panel prefixes

othercorey:
Fixes https://github.com/cakephp/debug_kit/issues/746

### DIFF
--- a/src/Panel/TimerPanel.php
+++ b/src/Panel/TimerPanel.php
@@ -33,7 +33,7 @@ class TimerPanel extends DebugPanel
     {
         $before = function ($name) {
             return function () use ($name) {
-                DebugTimer::start($name, __d('debug_kit', $name));
+                DebugTimer::start($name, $name);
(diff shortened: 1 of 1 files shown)


src/Panel/TimerPanel.php
```diff
@@ -61,18 +61,18 @@ public function implementedEvents(): array
                 ['priority' => 999, 'callable' => $after('Event: Controller.startup')],
                 ['priority' => 999, 'callable' => function () {
                     DebugMemory::record(__d('debug_kit', 'Controller action start'));
-                    DebugTimer::start(__d('debug_kit', 'Controller action'));
+                    DebugTimer::start(__d('debug_kit', 'Controller: action'));
```
ADmad, line 64: Could you remove the translation function usages too please. They don't serve any purpose here.

othercorey: done.

PR #751, opened 2020-04-13
Back-port #750 to 3.x

othercorey:
https://github.com/cakephp/debug_kit/pull/750

### DIFF
--- a/src/Panel/TimerPanel.php
+++ b/src/Panel/TimerPanel.php
@@ -32,7 +32,7 @@ class TimerPanel extends DebugPanel
     {
         $before = function ($name) {
             return function () use ($name) {
-                DebugTimer::start($name, __d('debug_kit', $name));
+                DebugTimer::start($name);
             };
         };
         $after = function ($name) {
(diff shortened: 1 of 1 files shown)


src/Panel/TimerPanel.php
```diff
@@ -59,19 +59,19 @@ public function implementedEvents()
...
-                    DebugTimer::stop(__d('debug_kit', 'Controller action'));
+                    DebugTimer::stop('Controller: action');
                 }],
                 ['priority' => 0, 'callable' => $before('Event: Controller.beforeRender')],
                 ['priority' => 999, 'callable' => $after('Event: Controller.beforeRender')],
                 ['priority' => 999, 'callable' => function () {
-                    DebugMemory::record(__d('debug_kit', 'View Render start'));
-                    DebugTimer::start(__d('debug_kit', 'View Render start'));
+                    DebugMemory::record('View Render start');
+                    DebugTimer::start('debug_kit', 'View: Render');
```
ADmad, line 74: `debug_kit` should not be the 1st arg.

othercorey: good catch.

PR #754, opened 2020-04-13
Fix CS errors

othercorey:


src/DebugSql.php
```diff
@@ -77,7 +77,9 @@ public static function sql(Query $query, $showValues = true, $showHtml = null, $
         }
 
         $trace = Debugger::trace(['start' => 1, 'depth' => $stackDepth + 2, 'format' => 'array']);
+        /** @psalm-suppress PossiblyInvalidArrayOffset */
         $file = isset($trace[$stackDepth]) ? $trace[$stackDepth]['file'] : 'n/a';
```
othercorey, line 81: I'm not really sure why psalm thinks these should be int offsets.

garas: This could be simplified to 
```suggestion
        $file = $trace[$stackDepth]['file'] ?? 'n/a';
```

PR #757, opened 2020-04-26
Port - Use rand() to trigger garbage collection

othercorey:
https://github.com/cakephp/debug_kit/pull/756


allanmcarvalho:
Captura de Tela 2020-05-01 às 18 45 28

I don't know why, but using the `dev-master` some times, in most of case, on first loads, when debug is on, I get this errors on some plugin assets files. And some times get the `database is locked` error yet!

othercorey:
Please open a separate issue.

Issue #759, opened 2020-05-08
Plugin assets conflicting with DebugKit


shahways:
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 4.0.7
* Platform and Target: 
DebugKit 4.1.2
macOS Catalina 10.15.4
MAMP with Apache 2.2.34,  PHP 7.4.2 and MySQL 5.7.26

### What you did
- Baked a theme-only plugin and added a 'templates' and 'webroot' folder. 
- Created a default.php layout and placed some image, css and js assets in the webroot folder of the plugin. 
- Activated the theme in the AppController in the beforeRender callback. 
- Created a view in the plugin in templates/Pages/test.php

### What happened
- The page keeps pending on the plugin assets (js/img/css). After a minute all the assets return a 500 error. 
- The error.log shows "Error: [PDOException] SQLSTATE[HY000]: General error: 5 database is locked", the stack trace shows 2 lines related to DebugKit (vendor/cakephp/debug_kit/src/ToolbarService.php:256 and vendor/cakephp/debug_kit/src/Middleware/DebugKitMiddleware.php:66)
- Loading a plugin asset directly through the URL works fine and fast
- Deleting the tmp/debug_kit.sqlite file had no effect
- When debug set to false, everything works perfectly

### What you expected to happen
When debugging is enabled, plugin assets should be loading just as fast without locking the DebugKit sqlite db. 

### Additional information
In cakephp/cakephp#757 (https://github.com/cakephp/debug_kit/pull/757) at the bottom @allanmcarvalho apparently describes the same situation, but I could not find any related issues here. His screenshot of his browser inspector looks exactly like mine.

PR #760, opened 2020-05-31
Allow saving requests to fail gracefully

othercorey:
Fixes https://github.com/cakephp/debug_kit/issues/759

We should allow concurrent requests without throwing fatal exceptions.

### DIFF
--- a/src/ToolbarService.php
+++ b/src/ToolbarService.php
@@ -22,6 +22,7 @@ use Cake\Log\Log;
 use Cake\ORM\TableRegistry;
 use Cake\Routing\Router;
 use DebugKit\Panel\PanelRegistry;
+use PDOException;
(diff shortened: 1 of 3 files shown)


src/ToolbarService.php
```diff
@@ -207,7 +208,7 @@ public function initializePanels()
      *
      * @param \Cake\Http\ServerRequest $request The request
      * @param \Psr\Http\Message\ResponseInterface $response The response
-     * @return null|\DebugKit\Model\Entity\Request Saved request data.
+     * @return false|\DebugKit\Model\Entity\Request Saved request data.
```
othercorey, line 211: `Table::save()` returns `false|EntityInterface`. Added false which was missing from the `return save()` and replaced null with false to avoid mixed types.

PR #761, opened 2020-05-31
Fixed cs errors. Added psalm composer scripts.

othercorey:


src/ToolbarService.php
```diff
@@ -145,7 +145,7 @@ protected function isSuspiciouslyProduction()
             Log::warning(
                 "DebugKit is disabling itself as your host `{$host}` " .
                 "is not in the known safe list of top-level-domains ({$safeList}). " .
-                "If you would like to force DebugKit on use the `DebugKit.forceEnable` Configure option."
+                'If you would like to force DebugKit on use the `DebugKit.forceEnable` Configure option.'
```
othercorey, line 148: I wonder if I can adjust the code sniff to now require changing mixed lines like this.

dereuromark: I wouldnt. Rather fix the others IMO if you want to harmonize it.

PR #764, opened 2020-06-08
Allow ignoring paths.

ADmad:
Refs #707

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -50,6 +50,12 @@ Configuration
     // Before loading DebugKit
     Configure::write('DebugKit.forceEnable', true);
 
+* ``DebugKit.ignorePaths`` - Regex pattern (including delimiter) to ignore paths.
+   DebugKit won't save data for request URLs which match this regex. Defaults to ``null``::
+
+    // Ignore image paths
+    Configure::write('DebugKit.ignorePaths', '/\.(jpg|png|gif)$/');
+
 * ``DebugKit.ignoreAuthorization`` - Set to true to ignore Cake Authorization plugin for DebugKit requests. Disabled by default.
 
 Database Configuration
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -51,7 +51,7 @@ Configuration
     Configure::write('DebugKit.forceEnable', true);
 
 * ``DebugKit.ignorePaths`` - Regex pattern (including delimiter) to ignore paths.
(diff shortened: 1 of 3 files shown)


markstory:
Looks good to me. What do you think @saeideng ?

docs/en/index.rst
```diff
@@ -50,6 +50,12 @@ Configuration
     // Before loading DebugKit
     Configure::write('DebugKit.forceEnable', true);
 
+* ``DebugKit.ignorePaths`` - Regex pattern (including delimiter) to ignore paths.
+   DebugKit won't save data for request URLs which match this regex. Defaults to ``null``::
```
othercorey, line 54:
```suggestion
   DebugKit won't save data for request URLs that match this regex. Defaults to ``null``::
```
https://www.grammarly.com/blog/which-vs-that/ (not to be pretentious, it's just a good explanation of why `which` feels wrong)

tests/TestCase/ToolbarServiceTest.php
```diff
@@ -164,6 +164,40 @@ public function testSaveDataIgnoreDebugKitDashedUrl()
         $this->assertFalse($bar->saveData($request, $response));
     }
 
+    /**
+     * Test that saveData ignores path which match "ignorePaths" regex.
```
othercorey, line 168:
```suggestion
     * Test that saveData ignores path that matches "ignorePaths" regex.
```
just being thorough

src/ToolbarService.php
```diff
@@ -67,6 +67,7 @@ class ToolbarService
         ],
         'forceEnable' => false,
         'safeTld' => [],
+        'ignorePaths' => null,
```
othercorey, line 70: Maybe this is cakephp convention, but `ignorePaths` doesn't feel like a pattern to me.

Up to you, it's not a major api.

docs/en/index.rst
ADmad: I actually appreciate the English grammar lesson :slightly_smiling_face:.

src/ToolbarService.php
ADmad: We could add a `Regex` suffix to the option name but I don't remember similar being done elsewhere.

othercorey: I was thinking `ignorePathsPattern` but whatever. If this is convention then it's convention.

saeideng:
> Looks good to me. What do you think @saeideng ?

I tested it locally, it works well 
good job @ADmad :+1:

Issue #771, opened 2020-08-05
Add variable type


gemal:
to be able to see if a variable is a int or string since it's impossible to see if a variable is "8". is that a string 8 or 8 int
today:
![image](https://user-images.githubusercontent.com/69105/89403707-079b6480-d719-11ea-886a-5c3aac8c3eec.png)

perhaps something like this:
![image](https://user-images.githubusercontent.com/69105/89403683-fe11fc80-d718-11ea-9592-36e6d0a51f88.png)

would that be something anybody would be interested in?

markstory:
I think this is already available in the 4.x version of debugkit. Were you hoping to see this change in an earlier release?

dereuromark:
I checked and didnt see it either, though.

markstory:
@dereuromark hrm. Maybe I only thought about doing it and then forgot to actually do it.

dereuromark:
![Screenshot from 2020-08-09 15-34-23](https://user-images.githubusercontent.com/39854/89733451-d5aa3b00-da55-11ea-855c-f78586db7f2b.png)

non-scalar ones seem to have that, and NULL
but not string/int etc, which are hard to tell the type of then.

markstory:
Would using the new debug output from 4.1 work to help solve this? I think it could make output more consistent with the rest of CakePHP, fix this problem and give us less code to maintain in the long run.  This is the change that I had thought I did but didn't actually do.

dereuromark:
PR opened for it.

dereuromark:
@markstory Using it now I discovered a few follow ups

The root level should probably stay the former list of expandable items
Right now the whole huge array is dumped at once
```
[
    '_authUser' => [
],
    'queuedJobs' => [
],
    'queuedJob' => [
...
```
which can be thousands and millions of rows directly

The other issue is the tick box "Sort variables by name" above that now also become not functional due to this.

markstory:
@dereuromark I think the collapsing is fixed in 4.1.3 I can verify that though. I'll check out the sort variables feature as that was done before the HTML was generated before.

PR #772, opened 2020-08-11
Feat/improve dump

markstory:
Make the variable dump output use the new HTML formatting provided in CakePHP. This makes DebugKit more consistent with the rest of Cake and will allow us to remove some code from DebugKit. Furthermore, it makes the types of variables easier to spot than today solving #771

### DIFF
--- a/templates/layout/panel.php
+++ b/templates/layout/panel.php
@@ -5,7 +5,7 @@
 echo $this->fetch('content');
 ?>
 <script type="text/javascript">
-if (window.__cakeDebugBlockInit()) {
+if (window.__cakeDebugBlockInit) {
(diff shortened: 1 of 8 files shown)


templates/layout/panel.php
```diff
@@ -2,5 +2,10 @@
 /**
  * @var \DebugKit\View\AjaxView $this
  */
-
 echo $this->fetch('content');
+?>
+<script type="text/javascript">
+if (window.__cakeDebugBlockInit()) {
```
othercorey, line 8:
```suggestion
if (window.__cakeDebugBlockInit) {
```

dereuromark:
Works like a charm :)

Issue #773, opened 2020-08-19
Notice: Undefined index: debugKitLog


dereuromark:
Latest 3.x version of DebugKit throws

> Notice (8): Undefined index: debugKitLog [ROOT/vendor/cakephp/debug_kit/src/Mailer/Transport/DebugKitTransport.php, line 37]

Maybe we can avoid this notice?

```
Cake\Core\BasePlugin::{closure}() - ROOT/vendor/cakephp/debug_kit/config/bootstrap.php, line 43
DebugKit\Mailer\Transport\DebugKitTransport::__construct() - ROOT/vendor/cakephp/debug_kit/src/Mailer/Transport/DebugKitTransport.php, line 37
DebugKit\Mailer\Transport\DebugKitTransport::__construct() - ROOT/vendor/cakephp/debug_kit/src/Mailer/Transport/DebugKitTransport.php, line 56
Cake\Mailer\TransportRegistry::_create() - CORE/src/Mailer/TransportRegistry.php, line 90
```

markstory:
Why is the constructor called twice in a row? Is DebugKit wrapping itself somehow?

dereuromark:
I had this case when upgrading an app from 3.x to 4.x
And it blocked moving forward with other things.
So it would be good if we find a solution to not through those warnings - as this complicates the upgrade path.

PR #774, opened 2020-08-19
Fix notice in transport

dereuromark:
Resolves https://github.com/cakephp/debug_kit/issues/773

### DIFF
--- a/src/Mailer/Transport/DebugKitTransport.php
+++ b/src/Mailer/Transport/DebugKitTransport.php
@@ -3,6 +3,7 @@ declare(strict_types=1);
 
 namespace DebugKit\Mailer\Transport;
 
+use ArrayObject;
 use Cake\Core\App;
 use Cake\Mailer\AbstractTransport;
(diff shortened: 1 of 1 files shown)


dereuromark:
If this is acceptable we should backport this to 3.x then (where this occurred)

src/Mailer/Transport/DebugKitTransport.php
```diff
@@ -36,7 +37,7 @@ class DebugKitTransport extends AbstractTransport
      */
     public function __construct($config = [], ?AbstractTransport $originalTransport = null)
     {
-        $this->emailLog = $config['debugKitLog'];
+        $this->emailLog = $config['debugKitLog'] ?? new ArrayObject();
```
markstory, line 40: Will this still capture email messages? I think the email panel relies on passing the log object in.

dereuromark:
I am done upgrading for now, so I guess I wont persue this further.

PR #775, opened 2020-08-20
Fix sorting and collapsing of dump output

### LINKED ISSUE #771: Add variable type

markstory:
I thought the panel layout was used. It is not.
I forgot to apply variable sorting when switching dump implementations.

Fixes #771


tests/TestCase/View/Helper/ToolbarHelperTest.php
```diff
@@ -78,10 +79,30 @@ public function testDumpCoerceHtml()
...
+
         // Restore back to original value.
         Debugger::configInstance('exportFormatter', $restore);
     }
 
+    public function testDumpSorted()
+    {
+        $getValues = function ($el) {
+            return (string)$el;
+        };
```
garas, line 91: Use `strval()` instead

markstory: ok.

dereuromark: Why that?

garas: Whole closure can be replaced with PHP function, not code in closure.

PR #779, opened 2020-08-25
sqlite3 extension

mhf-ir:


dereuromark:
This is not a requirement.
If sqlite is not present it will and can use a different configured DB, e.g. Mysql or Postgres.
It just needs "any" - you provided "none", that was the issue.

dereuromark:
It is quite clearly mentioned in the README install instructions, as well, already:
https://github.com/cakephp/debug_kit#requirements
So closing then, as nothing is needed to be fixed.

ndm2:
> [...] if you need to use a different database see the Database Configuration section below.

To be fair, there is no "_Database Configuration section_" below 😉  It should probably say something like "_see the Database Configuration section in the documentation linked below_", or directly link to it.

dereuromark:
Done in https://github.com/cakephp/debug_kit/commit/b72d5e4ab458e706e3452f49f8f9c65e3ae7e627

PR #782, opened 2020-10-08
fix: pass default-param to underlying cache-engine

### LINKED ISSUE #781: DebugEngine does not pass default value for getMultiple
DebugKit wraps Cache-Engines with the `DebugKit\Cache\Engine\DebugEngine` class.
This class reimplements the function `getMultiple` to track performance (i guess).
That function takes two parameters: `$keys` (the keys to get from cache) and `$default` (the value to return when there was no cache hit) and internally calls the `getMultiple` on the real engine ([see here](https://github.com/cakephp/debug_kit/blob/7e0de3676c867318d76c1293ff7fa8fbf075e1fc/src/Cache/Engine/DebugEngine.php#L199)).
However, the param `$default` is not passed to the call to the engine.

This breaks when using CakePHPs `Cache::pool` with `Psr16Adapter` from Symphony - which a library I'm using requires me to (it internally wraps a given pool).
This breaks because said Adapter in Symphony internally uses a reference to `\stdObject` to determine whether the cache had a hit or not (which allows them to store `null`-Values in the Cache). And thus, cache-misses are not recognised and null is returned as the cached value.

Said `Psr16Adapter` requires the given pool to implement `Psr\SimpleCache\CacheInterface`, as which the return-value of CakePHPs `Cache::pool` is defined (as well as the class `Cake\Cache\CacheEngine`, which said `DebugEngine` extends).

Is there some intention why this param is not passed?
If not I'm happy to send a pull-request to fix this issue (if you'd like one).

xyng:
fixes #781

### DIFF
--- a/src/Cache/Engine/DebugEngine.php
+++ b/src/Cache/Engine/DebugEngine.php
@@ -196,7 +196,7 @@ class DebugEngine extends CacheEngine
     public function getMultiple($keys, $default = null): iterable
     {
         $start = microtime(true);
-        $result = $this->_engine->getMultiple($keys);
+        $result = $this->_engine->getMultiple($keys, $default);
         $duration = microtime(true) - $start;
 
         $this->track('get hit');


othercorey:
Thanks!

PR #783, opened 2020-10-11
fix warning in php8

swiffer:
DebugKit\Panel\VariablesPanel::DebugKit\Panel\{closure}(): Argument #1 ($item) must be passed by reference, value given


markstory:
Thank you 👍 I'll look into the build failure as it doesn't seem like it should be related.

hackzilla:
@markstory Any chance this could be merged into the 3.x branch as well?
I'm having this error with cake 3.9 & 3.10 ~~with php 7.3 and php 7.4~~

Edit: CakePHP is defaulting to PHP 8.0 when I run the server command.

PR #784, opened 2020-10-12
Add php8 build

markstory:
Improve test failure message and add php8 coverage to try and diagnose the root cause of errors in php8 that may have been causing #783


markstory:
Closing as we're getting off of travis.

PR #790, opened 2020-11-07
Add Github Actions based CI

markstory:
Convert what we had in travis to GitHub Actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,10 @@ jobs:
       fail-fast: false
       matrix:
         php-version: ['7.2', '7.4']
-        db-type: [mysql, pgsql]
+        db-type: [mysql, pgsql, sqlite]
         prefer-lowest: ['']
         include:
-          - php-version: '7.2'
+          - php-version: '7.4'
             db-type: 'sqlite'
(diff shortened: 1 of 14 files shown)


composer.json
```diff
@@ -32,7 +32,8 @@
     "require-dev": {
         "cakephp/cakephp-codesniffer": "^4.0",
         "cakephp/authorization": "^2.0",
-        "phpunit/phpunit": "^8.0"
+        "phpunit/phpunit": "^8.0",
+        "psalm/phar": "^3.6"
```
othercorey, line 36: This should match the version used in GA. We should remove psalm-setup if we're adding it here.

markstory: Whoops this isn't supposed to be here.

.github/workflows/ci.yml
```diff
@@ -15,10 +15,10 @@ jobs:
       fail-fast: false
       matrix:
         php-version: ['7.2', '7.4']
-        db-type: [mysql, pgsql]
+        db-type: [mysql, pgsql, sqlite]
         prefer-lowest: ['']
         include:
-          - php-version: '7.2'
+          - php-version: '7.4'
```
othercorey, line 21: prefer-lowest should be 7.2 not 7.4

PR #791, opened 2020-11-10
Remove travis and add php8 support

markstory:


othercorey:
The extra phpunit methods were added to `TestCase` to be backwards compatible. Will need to do the same here somehow:

https://github.com/cakephp/cakephp/commit/e01649deb97ec7749cae0d900be18ff6d6859b09#diff-b24fba393b9e95eaea01b9206716931a735e88744f97b9f9d5d9f30889bc7fba

markstory:
@othercorey I'll just revisit PHP8 support after 4.2 comes out

othercorey:
Do we need to test more than sqlite with this one or should we add windows build for the sqlserver specific tests?

markstory:
> Do we need to test more than sqlite with this one or should we add windows build for the sqlserver specific tests?

I don't think there is a need for a sqlserver build. There isn't any logic in this plugin that uses advanced database features.

Issue #793, opened 2020-11-15
4.3.6 excludes /tests from archive so composer install --prefer-dist will have broken debug_kit (Fixture missing)


swiffer:
Fatal error: Uncaught Error: Class name must be a valid object or a string in /var/www/html/tiefblau/vendor/cakephp/debug_kit/src/Model/Table/LazyTableTrait.php on line 64

othercorey:
Thanks for finding and reporting this. Should be fixed in 4.3.7 release.

PR #794, opened 2020-11-15
Update .gitattributes

swiffer:
Keep Fixtures so DebugKit is able to create Tables when required (LazyTableTrait).

solves #793 , recovers #792


othercorey:
Thanks. It's unfortunate that this is part of tests, but we definitely need it.

PR #796, opened 2020-11-15
WIP: Fix serialization in Variables Panel with PHP8

swiffer:
array_map can not be used with value passed by reference, so use array_walk instead ?


othercorey:
Fixed in separate PR.

PR #804, opened 2020-11-28
Fix deprecations

othercorey:


othercorey:
Sometimes gc will run the first time it is called when there are no requests which throws an expression error.

PR #808, opened 2020-12-02
Improve variables panel

### LINKED ISSUE #277: Output of variables tab makes relations of data hard to spot.
debugkit

It is very hard to spot the levels of nesting and the relations of the data inside the variables in 3.x.
I missed the second level (event) completely twice, for example:

```
$attendee->event->propery
```

Is there a way to make those nestings and relations of data (Attendee belongsTo Event in this case) clearer here?
Maybe highlight entities with a light yellow?

### LINKED ISSUE #659: Variables: Lying array type
Two things:
Larger view template data passed down makes the view rendering super slow as debug kit always renders out ALL variables incl nesting.
It should, ideally, lazyload those if the panel is opened. Or even per layer if clicked upon. Not all at once even with closed debug kit panels.

Then, the types are also lying :
toArray() seems to eagertransform the entity incl all nesting entities. So the actual class names are gone then. Instead of an (array) I would expect (FooEntity) displayed to know the real type that was there.

Any idea how we could achieve a better debug kit result here?

### LINKED ISSUE #795: VariablesPanel: Passed by reference Error in PHP 8
I'm getting a lot of these (testing with PHP8.0RC3)

Warning (2): DebugKit\Panel\VariablesPanel::DebugKit\Panel\{closure}(): Argument #1 ($item) must be passed by reference, value given [ROOT/vendor/cakephp/debug_kit/src/Panel/VariablesPanel.php, line 86]

markstory:
Use the new Debugger features to generate a more complete dataset for
the Variables Panel. The new data will include hidden properties and
not mangle Entity data.

Fixes #795
Fixes #659
Fixes #277

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require": {
         "php": ">=7.2",
-        "cakephp/cakephp": "^4.2.0-RC1",
+        "cakephp/cakephp": "^4.2@RC",
         "cakephp/chronos": "^2.0",
         "composer/composer": "^1.3 | ^2.0",
         "jdorn/sql-formatter": "^1.2"
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require": {
         "php": ">=7.2",
-        "cakephp/cakephp": "^4.1",
+        "cakephp/cakephp": "^4.2.0-RC1",
         "cakephp/chronos": "^2.0",
         "composer/composer": "^1.3 | ^2.0",
         "jdorn/sql-formatter": "^1.2"
--- a/src/View/Helper/ToolbarHelper.php
+++ b/src/View/Helper/ToolbarHelper.php
@@ -16,7 +16,10 @@ declare(strict_types=1);
 namespace DebugKit\View\Helper;
 
 use ArrayAccess;
+use Cake\Error\Debug\ArrayItemNode;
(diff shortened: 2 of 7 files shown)


othercorey:
Need to add php 8.0 to GA matrix.

src/View/Helper/ToolbarHelper.php
```diff
@@ -57,11 +60,38 @@ public function setSort($sort)
...
     }
 
+    /**
+     * Dump an array of nodes
+     *
+     * @param \Cake\Error\Debug\NodeInterface[] $nodes An array of dumped variables.
+     *   Variables should be keyed by the name they had in the view.
+     * @return string Formatted HTML
+     */
+    public function dumpNodes($nodes)
```
othercorey, line 70:
```suggestion
    public function dumpNodes(array $nodes): string
```

The code below has an `in_array()` check but the param doc say it's an array. Should always be an array, right?

composer.json
```diff
@@ -24,7 +24,7 @@
     },
     "require": {
         "php": ">=7.2",
-        "cakephp/cakephp": "^4.1",
+        "cakephp/cakephp": "^4.2.0-RC1",
```
ADmad, line 27:
```suggestion
        "cakephp/cakephp": "^4.2@RC",
```
Looks cleaner :)

PR #810, opened 2020-12-21
Reduce maximum node depth for variables to 3 to avoid massive memory use

othercorey:


othercorey:
Can we add this depth to the DebugKit configure and read it here?

markstory:
5 is probably ok.

PR #825, opened 2021-05-08
Fix failure when debug_kit connection does not exist 

markstory:
When the host application doesn't have SQLite installed failing to save the toolbar data should not cause the page to 500. Instead we can log the error and not end up showing the toolbar iframe as there is not Request persisted.

Refs cakephp/cakephp#15492

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -7,7 +7,7 @@
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
     autoloader="tests/bootstrap.php"
     usePhpDocMethodsWithoutMagicCall="true"
-    errorBaseline="psalm-baseline.xml"
+    errorBaseline="./psalm-baseline.xml"
 >
     <projectFiles>
         <directory name="src" />
--- a/src/ToolbarService.php
+++ b/src/ToolbarService.php
@@ -260,7 +260,10 @@ class ToolbarService
             $requests = $this->getTableLocator()->get('DebugKit.Requests');
             $requests->gc();
         } catch (MissingDatasourceConfigException $e) {
-            Log::warning('Unable to save request. Check your custom debut_kit datasource connection or check that SQLite extension is enabled..');
+            Log::warning(
+                'Unable to save request. Check your debug_kit datasource connection ' .
+                'or ensure that PDO SQLite extension is enabled.'
+            );
             Log::warning($e->getMessage());
 
             return false;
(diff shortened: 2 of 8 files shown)


src/ToolbarService.php
```diff
@@ -254,9 +255,16 @@ public function saveData(ServerRequest $request, ResponseInterface $response)
...
         ];
-        /** @var \DebugKit\Model\Table\RequestsTable $requests */
-        $requests = $this->getTableLocator()->get('DebugKit.Requests');
-        $requests->gc();
+        try {
+            /** @var \DebugKit\Model\Table\RequestsTable $requests */
+            $requests = $this->getTableLocator()->get('DebugKit.Requests');
+            $requests->gc();
+        } catch (MissingDatasourceConfigException $e) {
+            Log::warning('Unable to save request. You need to define the debug_kit connection.');
```
othercorey, line 263:
```suggestion
            Log::warning('Unable to save request. Check your custom debut_kit database connection or check that SQLite extension is enabled..');
```

src/ToolbarService.php
```diff
@@ -254,9 +255,19 @@ public function saveData(ServerRequest $request, ResponseInterface $response)
...
-        /** @var \DebugKit\Model\Table\RequestsTable $requests */
-        $requests = $this->getTableLocator()->get('DebugKit.Requests');
-        $requests->gc();
+        try {
+            /** @var \DebugKit\Model\Table\RequestsTable $requests */
+            $requests = $this->getTableLocator()->get('DebugKit.Requests');
+            $requests->gc();
+        } catch (MissingDatasourceConfigException $e) {
+            Log::warning(
+                'Unable to save request. Check your debug_kit datasource connection ' .
```
ADmad, line 264: We should also add a big fat note in the readme asking people to check the error logs for details if Debugkit isn't showing up :).

markstory: Done. We can always add more reasons in the future.

psalm.xml
```diff
@@ -7,7 +7,7 @@
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
     autoloader="tests/bootstrap.php"
     usePhpDocMethodsWithoutMagicCall="true"
-    errorBaseline="psalm-baseline.xml"
+    errorBaseline="./psalm-baseline.xml"
```
dereuromark, line 10: Is this really necessary?

PR #827, opened 2021-05-14
Update composer requirement to allow 1.x or 2.x

fortkle:
Update composer requirement to allow 1.x or 2.x. 
This change is a backport of #786 to debug_kit:3.x branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "cakephp/cakephp": "^3.7.0",
         "cakephp/chronos": "^1.0.0",
         "cakephp/plugin-installer": "^1.0.0",
-        "composer/composer": "^1.3.0",
+        "composer/composer": "^1.3 | ^2.0",
         "jdorn/sql-formatter": "^1.2.0"
     },
     "require-dev": {


othercorey:
The 3.x branch only has a travis config so unit tests didn't run.

fortkle:
@othercorey 
Thank you for checking!

> The 3.x branch only has a travis config so unit tests didn't run.

Oh, I see. Is there anything I can do to help?
I want to use composer scripts with composer v2, but I'm having trouble with CakePHP3 because I can only use composer v1 due to the cakephp/debug_kit dependency.
I will be happy if composer v2 can be used as well.

I'll spare you the details, but there is a part in the [composer script](https://getcomposer.org/doc/articles/scripts.md) that executes the composer command, and there is a problem where the PATH resolution does not work and the v1 binary in the vendor directory is referenced.

markstory:
I can get CI configuration backported.

othercorey:
Merged in https://github.com/cakephp/debug_kit/pull/828

fortkle:
Thank you! 👍 @othercorey @markstory

PR #835, opened 2021-09-02
Show Toolbar on Dashboard - Reload Request History

### LINKED ISSUE #834: Show history and related panels in dashboard
Related to "Last requests to inspect in detail maybe" in #554

I'm using a cakephp mainly as a rest-only backend, so i don't see much the toolbar. I could really use the panels to inspect each request made by the front-end (a separate non-php project) or mobile apps.

A quick way to enable this would be just show the toolbar in the dashboard, and a button in the requests panel to reload the requests (or refresh the page). If you are ok with this I can try open a PR.

raul338:
Closes #834

With the added Reload button, the "back to current request" will never be the same. Should I add some logic to keep the request id ?

Also, any hint about how should I test this will be appreciated

### DIFF
--- a/src/Controller/PanelsController.php
+++ b/src/Controller/PanelsController.php
@@ -104,7 +104,7 @@ class PanelsController extends DebugKitController
             ->disableHydration()
             ->first();
         if (!$request) {
-            throw new NotFoundException();
+            throw new NotFoundException('No requests found');
         }
         /** @var array{id:string}|null $historyPanel */
         $historyPanel = $this->Panels->find('byRequest', ['requestId' => $request['id']])
@@ -112,7 +112,7 @@ class PanelsController extends DebugKitController
             ->select(['id'])
             ->first();
         if (!$historyPanel) {
-            throw new NotFoundException();
+            throw new NotFoundException('History Panel from latest request not found');
         }
 
         return $this->redirect([
--- a/src/Controller/PanelsController.php
+++ b/src/Controller/PanelsController.php
@@ -83,11 +83,7 @@ class PanelsController extends DebugKitController
     public function view($id = null)
     {
         $this->set('sort', $this->request->getCookie('debugKit_sort'));
-        $panel = $this->Panels->get($id);
-        if ($panel->title === 'History') {
-            $request = $this->Panels->Requests->get($panel->request_id);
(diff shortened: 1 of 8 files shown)


src/Controller/PanelsController.php
```diff
@@ -84,10 +84,46 @@ public function view($id = null)
     {
         $this->set('sort', $this->request->getCookie('debugKit_sort'));
         $panel = $this->Panels->get($id);
+        if ($panel->title === 'History') {
+            $request = $this->Panels->Requests->get($panel->request_id);
+            $this->set('currentRequest', $request);
+        }
```
markstory, line 90: Couldn't this be a property that the History panel provides as its `data()` results?

raul338: It will be required to save the request entity separate from the panels. So that way it will show in the `recent` finder.
If its ok, i think I can do that.

markstory: Hrm. Could the Request be loaded with `contain()` That would allow `panel->request` to always contain the related request for the panel.

raul338: That's better. Done!

markstory:
Looks like there is a test failure with SQLite.

raul338:
The pass locally (and fails on others request) and I'm using in my demo project with sqlite and works. no idea why it fails.  I added a message to the exception to see which fails

templates/element/history_panel.php
```diff
@@ -20,16 +19,25 @@
 ?>
 <div id="request-history">
 <?php if (empty($requests)): ?>
-    <p class="warning"><?= __d('debug_kit', 'No previous requests logged.') ?></p>
+    <p class="warning">
+        <?= __d('debug_kit', 'No requests logged.') ?>
+        <button type="button" onclick="toolbar.loadPanel('latest-history')"><?= __d('debug_kit', 'Reload') ?></button>
+    </p>
```
raul338, line 25: On reload, they are no longer "previous" requests.

markstory:
Looking good, but there is a failing test.

raul338:
> Looking good, but there is a failing test.

For some reason, `$this->makeRequest()` is not the most recent request :shrug:

raul338:
Is there anything else to do here?

PR #836, opened 2021-09-02
Add php 8.1 to test matrix

othercorey:


othercorey:
Looks like it will pass once 4.3 is released.

PR #839, opened 2021-09-14
Fix deprecation warnings.

ADmad:


tests/TestCase/Panel/VariablesPanelTest.php
```diff
@@ -74,9 +74,9 @@ public function testShutdown()
 
         $unserializableDebugInfo = $this
             ->getMockBuilder('\stdClass')
-            ->setMethods(['__debugInfo'])
+            ->addMethods(['__debugInfo'])
             ->getMock();
-        $unserializableDebugInfo->expects($this->once())->method('__debugInfo')->willReturn([
+        $unserializableDebugInfo->expects($this->any())->method('__debugInfo')->willReturn([
```
ADmad, line 79: @markstory This test was failing as `__debugInfo()` was being called more than once. I couldn't figure out how/why that was happening. Most likely related to the Debugger internals changes.

ADmad:
The remaining failures are probably due to tables not being dropped/truncated between tests with the new fixture manager.

tests/TestCase/Panel/SqlLogPanelTest.php
```diff
@@ -25,13 +25,6 @@
  */
 class SqlLogPanelTest extends TestCase
 {
-    /**
-     * fixtures.
-     *
-     * @var array
-     */
-    public $fixtures = ['core.Articles'];
```
markstory, line 33: Do we need the records in this fixture?

tests/TestCase/Panel/VariablesPanelTest.php
markstory: I suspect you're right. Having the mock always return the same value for __debugInfo should be fine.

tests/schema.php
```diff
@@ -0,0 +1,61 @@
...
+declare(strict_types=1);
+
+use Cake\Database\Schema\TableSchema;
+
+/**
+ * Abstract schema for Debugkit tests.
+ */
+return [
+    [
+        'table' => 'panels',
```
markstory, line 11: I wonder if we should also update the schema generation tricks we use in DebugKit to build the database at runtime to also use this schema format.

markstory (commented):
Thanks for getting this started. I can take care of getting the rest of the tests passing if you'd like.

ADmad:
> I can take care of getting the rest of the tests passing if you'd like.

Yes please.

tests/TestCase/Panel/SqlLogPanelTest.php
ADmad: Yes there are tests using Articles table/fixture.

PR #843, opened 2021-10-07
ModelAwareTrait is depreacated in 4.3

swiffer:


markstory:
I think this might be covered by #839

aa-matthias:
Hi @markstory - #839 is merged already in 4.next so i guess it's not covered.

ADmad said it's not neccessary here: https://github.com/cakephp/debug_kit/pull/844#issuecomment-937762011

Fine with it, I added the LocatorAwareTrait to the class myself where needed.

swiffer:
#839 is merged already in 4.next so i guess it's not covered.

ADmad said it's not neccessary here: https://github.com/cakephp/debug_kit/pull/844#issuecomment-937762011

Fine with it, I added the LocatorAwareTrait to the class myself where needed.

ADmad:
> Fine with it, I added the LocatorAwareTrait to the class myself where needed.

@swiffer Can you share how do you use a model instance in your mailer preview class? I thought it would be uncommon since the mailer classes themselves have access to the modelawaretrait/locatorawaretrait. I personally haven't used the preview class so am unaware how people use it :)

swiffer:
sure - so in a UserMailPreview I'm doing something like this:

```php
        $this->fetchTable('Users');

        /** @var \App\Model\Entity\User|null $user */
        $user = $this->Users->find('all')->first();
        $user->reset_password_token = 'dummy-token';

        /** @var \App\Mailer\UserMailer $mailer */
        $mailer = $this->getMailer('User');

        return $mailer
            ->resetPassword($user);
```

there are other options - not passing the complete entity into the mailer or creating a new user instead of loading the first one from db but for now it's working fine for me

dereuromark:
fetchTable() doesnt auto assign anymore, as the code shows (if you open it up)
Instead, it is expected to use

    $user = $this->fetchTable('Users')->find()
    
etc

swiffer:
@dereuromark - thanks, you're completly right - it must be

```php
        $usersTable = $this->fetchTable('Users');

        /** @var \App\Model\Entity\User|null $user */
        $user = $usersTable->find('all')->first();
```

ADmad:
We can add `LocatorAwareTrait` to `MailPreview` if such usage is common. Guess I was mistaken about the usage of model instances in `MailPreview`.

dereuromark:
Cannot they add the trait themselves? This one liner seems easy enough.

PR #844, opened 2021-10-07
Fix deprecation errors.

ADmad:


ADmad:
`MailPreview` uses `ModelAwareTrait` which I have kept for BC. I haven't added `LocatorAwareTrait` as I don't think it's necessary and users can always add it your their preview class if needed.

src/Controller/DashboardController.php
```diff
@@ -44,11 +44,11 @@ public function beforeFilter(EventInterface $event)
      */
     public function index()
     {
-        $this->loadModel('DebugKit.Requests');
+        $requestsModel = $this->fetchTable('DebugKit.Requests');
```
dereuromark, line 47: You can only do this if you set the constraint to ^4.3 on CakePHP
The downside is that people wanting to have bugfixes for 4.2 will be blocked out then etc.

Thats why I prefer not to use new 4.3 stuff in plugins until no other options exist anymore, to avoid shutting people out that didn't have the change to upgrade yet. At least support for one minor behind should always be possible for most plugins.

dereuromark: Oh, I see we are already doing that:
https://github.com/cakephp/debug_kit/blob/4.next/composer.json#L27

src/Panel/PanelRegistry.php
```diff
@@ -79,6 +79,7 @@ protected function _throwMissingClassError(string $class, ?string $plugin): void
      */
     protected function _create($class, string $alias, array $config)
     {
+        /** @var \DebugKit\DebugPanel $instance */
```
dereuromark: Would this be not necessary if we adjust above psalm param as

    @psalm-param class-string<\DebugKit\DebugPanel> $class
    
?

PR #846, opened 2021-11-27
add ability to change variables panel max depth via config + add notice in panel

LordSimal:
image

Someone recently in the Slack support channel thought that his associations were not working.

After some time we found out, that he only looked into the variables panel of the debug kit but never outputted his variable via e.g. `pr()` or `var_dump()`.

He was just expecting that the variables tab shows all available data (including theoretically infinite associations) which is not the case.

But the panel doesn't inform users about that limitation. Also this limitation was "hardcoded" in the plugin and therefore couldn't be increased (if e.g. one has more resources available and really needs it)

Therefore this PR

* adds the ability to overwrite the default 5 levels of depth via the Config key `DebugKit.variablesPanelMaxDepth` and
* adds a notice in the variables tab to show the current amount of max levels + how the user can overwrite it if he wants to

### DIFF
--- a/src/Panel/VariablesPanel.php
+++ b/src/Panel/VariablesPanel.php
@@ -92,7 +92,7 @@ class VariablesPanel extends DebugPanel
         $errors = [];
         $content = [];
         $vars = $controller->viewBuilder()->getVars();
-        $varsMaxDepth = (int)Configure::read('DebugKit.variablesPanelMaxDepth') ?: 5;
+        $varsMaxDepth = (int)Configure::read('DebugKit.variablesPanelMaxDepth', 5);
 
         foreach ($vars as $k => $v) {
             // Get the validation errors for Entity
--- a/src/Panel/VariablesPanel.php
+++ b/src/Panel/VariablesPanel.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
  */
 namespace DebugKit\Panel;
 
+use Cake\Core\Configure;
 use Cake\Datasource\EntityInterface;
 use Cake\Error\Debugger;
 use Cake\Event\EventInterface;
@@ -91,6 +92,7 @@ class VariablesPanel extends DebugPanel
         $errors = [];
         $content = [];
         $vars = $controller->viewBuilder()->getVars();
+        $varsMaxDepth = (int)Configure::read('DebugKit.variablesPanelMaxDepth') ?: 5;
 
         foreach ($vars as $k => $v) {
             // Get the validation errors for Entity
@@ -102,12 +104,13 @@ class VariablesPanel extends DebugPanel
                     $errors[$k] = $formErrors;
                 }
             }
-            $content[$k] = Debugger::exportVarAsNodes($v, 5);
+            $content[$k] = Debugger::exportVarAsNodes($v, $varsMaxDepth);
         }
 
         $this->_data = [
             'variables' => $content,
             'errors' => $errors,
+            'varsMaxDepth' => $varsMaxDepth,
         ];
     }
 
--- a/templates/element/variables_panel.php
+++ b/templates/element/variables_panel.php
@@ -24,7 +24,7 @@ if (isset($error)) :
     printf('<p class="warning">%s</p>', $error);
 endif;
 
-if(isset($varsMaxDepth)){
+if (isset($varsMaxDepth)) {
     printf('<p class="info">%s</p>', sprintf(__d('debug_kit', 'This view does not show very deep associations! Currently %s levels are being shown. You can overwrite this via the config key %s'), $varsMaxDepth, '<strong>DebugKit.variablesPanelMaxDepth</strong>'));
 }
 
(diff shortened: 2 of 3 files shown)


LordSimal:
I am happy to adjust the text or any documentation after it has been confirmed, that this is acceptable 😄

saeideng:
by current panel design , showing a lot data could cause it crashes
I prefer to only show a message something like ` max depth is `X`  , for deep debug  use 'debug()' /  .... `

src/Panel/VariablesPanel.php
```diff
@@ -91,6 +92,7 @@ public function shutdown(EventInterface $event)
         $errors = [];
         $content = [];
         $vars = $controller->viewBuilder()->getVars();
+        $varsMaxDepth = (int)Configure::read('DebugKit.variablesPanelMaxDepth') ?: 5;
```
ADmad:
```suggestion
        $varsMaxDepth = (int)Configure::read('DebugKit.variablesPanelMaxDepth', 5);
```

templates/element/variables_panel.php
```diff
@@ -24,6 +24,10 @@
     printf('<p class="warning">%s</p>', $error);
 endif;
 
+if (isset($varsMaxDepth)) {
+    printf('<p class="info">%s</p>', sprintf(__d('debug_kit', 'This view does not show very deep associations! Currently %s levels are being shown. You can overwrite this via the config key %s'), $varsMaxDepth, '<strong>DebugKit.variablesPanelMaxDepth</strong>'));
```
ADmad: The message is a bit tool long. It can just state the current max depth (no reference of associations is necessary) and how to configure it along with a warning that higher depth can lead to running out of memory.

ADmad:
@saeideng Making the depth configurable allows one to use a lower value too :)

templates/element/variables_panel.php
LordSimal: The 120 character limit per line is killing me here 😂

LordSimal:
The current version looks like this
image

templates/element/variables_panel.php
othercorey: I think the phrase `nested arrays` would work.

LordSimal:
Newest version
image

templates/element/variables_panel.php
```diff
@@ -24,6 +24,14 @@
     printf('<p class="warning">%s</p>', $error);
 endif;
 
+if (isset($varsMaxDepth)) {
+    $msg = sprintf(__d('debug_kit', '%s levels of nested arrays shown.'), $varsMaxDepth);
```
markstory:
```suggestion
    $msg = sprintf(__d('debug_kit', '%s levels of nested data shown.'), $varsMaxDepth);
```

More than arrays gets limited.

templates/element/variables_panel.php
```diff
@@ -24,6 +24,14 @@
...
+    $msg .= __d('debug_kit', 'You can overwrite this via the config key');
+    $msg .= ' <strong>DebugKit.variablesPanelMaxDepth</strong>';
+    $msg .= __d('debug_kit', ' | Be aware that increasing the level depth can lead to an out of memory error.');
```
markstory:
```suggestion
    $msg .= __d('debug_kit', ' | Be aware that increasing the depth value can lead to an out of memory error.');
```

othercorey: This can be simpler: Increasing the depth value can lead to an out of memory error.

LordSimal: I now used the version from @othercorey

templates/element/variables_panel.php
```diff
@@ -24,6 +24,14 @@
     printf('<p class="warning">%s</p>', $error);
 endif;
 
+if (isset($varsMaxDepth)) {
+    $msg = sprintf(__d('debug_kit', '%s levels of nested data shown.'), $varsMaxDepth);
+    $msg .= __d('debug_kit', 'You can overwrite this via the config key');
+    $msg .= ' <strong>DebugKit.variablesPanelMaxDepth</strong>';
+    $msg .= __d('debug_kit', ' | Increasing the depth value can lead to an out of memory error.');
```
ADmad: Having the warning on a new line instead would provide better visibility/readability.

LordSimal: So like that?

image

ADmad: That takes up lot of space. Perhaps keep it in the `info` itself using `<br>`. Also a space is need before `You can...`.

LordSimal: I would have to adjust the styling of the info class because it is currently not meant to have multiline text.

image

ADmad: Pretty please :)

LordSimal: Would look like this and will adjust accordingly to how many lines are present.

image

LordSimal: image

PR #851, opened 2021-12-11
add ability to filter plugin routes in routes tab

### LINKED ISSUE #850: RFC: Group plugin-routes into tab-sections in routes panel
I just thought of a UX improvement for the routes tab: In my case I have 100 loaded routes.

image

These are not sorted or grouped in any particular way (inverse route key order as I have seen in the code) but still a bit "messy" I would say.

My improvement would be to introduce tabs at the top similar to the already present "Toggle debugkit internal routes" button.
The amount of tabs is depending on the available plugins / loaded routes in those plugins.

image

This is a very rude, quick mockup I just made but I guess you get the point.

Clicking on one of these buttons will show only the routes loaded for that particular plugin. 

What are your opinions about that?

LordSimal:
This PR refactors the routes tab in that sense that you are now able to filter routes per plugin.

Also you can see how many routes are loaded for that specific plugin next to the filter button.

Activating the button will **hide** the routes for that selected plugin.

The DebugKit routes are hidden by default as they have been before as well.

Fixes #850

### DIFF
--- a/templates/element/routes_panel.php
+++ b/templates/element/routes_panel.php
@@ -33,10 +33,13 @@ foreach (CorePlugin::loaded() as $plugin_name) {
 ?>
 <div class="debugkit-plugin-routes-button-wrapper">
     <button type="button" class="btn-primary js-debugkit-toggle-plugin-route" data-plugin=".route-entry--app">
-        <?= __d('debug_kit', 'App') ?><?= (!empty($amount_of_routes_per_group['app'])) ? ' (' . $amount_of_routes_per_group['app'] . ')' : '' ?>
+        <?= __d('debug_kit', 'App') ?>
+        <?= (!empty($amount_of_routes_per_group['app'])) ? ' (' . $amount_of_routes_per_group['app'] . ')' : '' ?>
     </button>
     <?php foreach($plugin_names as $plugin_name => $parsed_name): ?>
-        <button type="button" class="btn-primary js-debugkit-toggle-plugin-route<?= (strpos($plugin_name, 'DebugKit') === 0) ? ' is-active' : '' ?>" data-plugin=".route-entry--plugin-<?= $parsed_name ?>">
+        <button type="button" class="btn-primary js-debugkit-toggle-plugin-route
+            <?= (strpos($plugin_name, 'DebugKit') === 0) ? ' is-active' : '' ?>"
+            data-plugin=".route-entry--plugin-<?= $parsed_name ?>">
             <?= $plugin_name ?>
         </button>
     <?php endforeach; ?>
@@ -56,7 +59,8 @@ foreach (CorePlugin::loaded() as $plugin_name) {
         if (empty($route->defaults['plugin'])):
             $class = 'route-entry route-entry--app';
         else:
-            $class = 'route-entry route-entry--plugin route-entry--plugin-' . preg_replace('/\W+/','',strtolower($route->defaults['plugin']));
+            $class = 'route-entry route-entry--plugin route-entry--plugin-' .
+                preg_replace('/\W+/','',strtolower($route->defaults['plugin']));
 
             // Hide DebugKit internal routes by default
             if($route->defaults['plugin'] === 'DebugKit') $class .= ' hidden';
--- a/templates/element/routes_panel.php
+++ b/templates/element/routes_panel.php
@@ -6,6 +6,7 @@
 
 use Cake\Core\Plugin as CorePlugin;
 use Cake\Utility\Hash;
+use Cake\Utility\Text;
 
 $routes = Cake\Routing\Router::routes();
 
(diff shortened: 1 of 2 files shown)


templates/element/routes_panel.php
```diff
@@ -3,13 +3,46 @@
  * @var \Cake\Routing\Route\Route[] $routes
  * @var string $matchedRoute
  */
+
+use Cake\Core\Plugin as CorePlugin;
 use Cake\Utility\Hash;
 
 $routes = Cake\Routing\Router::routes();
+
+$amount_of_routes_per_group = [];
```
ADmad: We conventionally use `$camelBacked` var names.

LordSimal:
Stickler-CI doesn't like my way of adding styling to the `is-active` filter buttons 😠

templates/element/routes_panel.php
LordSimal: converted that now 👍🏻

templates/element/routes_panel.php
```diff
@@ -3,13 +3,49 @@
...
+    } else {
+        if (!array_key_exists($route->defaults['plugin'], $amountOfRoutesPerGroup)) {
+            $amountOfRoutesPerGroup[$route->defaults['plugin']] = 0;
+        }
+        $amountOfRoutesPerGroup[$route->defaults['plugin']]++;
+    }
+}
+
+$plugin_names = [];
+foreach (CorePlugin::loaded() as $plugin_name) {
```
dereuromark: All the var names are wrong, see also below.

LordSimal: god damn I am so used to that 😅

LordSimal: I think I got everything now

templates/element/routes_panel.php
```diff
@@ -3,13 +3,49 @@
...
+        if (!array_key_exists('app', $amountOfRoutesPerGroup)) {
+            $amountOfRoutesPerGroup['app'] = 0;
+        }
+        $amountOfRoutesPerGroup['app']++;
+    } else {
+        if (!array_key_exists($route->defaults['plugin'], $amountOfRoutesPerGroup)) {
+            $amountOfRoutesPerGroup[$route->defaults['plugin']] = 0;
+        }
+        $amountOfRoutesPerGroup[$route->defaults['plugin']]++;
+    }
```
garas:
```suggestion
    $group = $route->defaults['plugin'] ?? 'app';
    if (!array_key_exists($group, $amountOfRoutesPerGroup)) {
        $amountOfRoutesPerGroup[$group] = 0;
    }
    $amountOfRoutesPerGroup[$group]++;
```

LordSimal: thanks 🙇🏻

templates/element/routes_panel.php
```diff
@@ -3,13 +3,43 @@
...
+        $amountOfRoutesPerGroup[$group] = 0;
+    }
+    $amountOfRoutesPerGroup[$group]++;
+}
+
+$pluginNames = [];
+foreach (CorePlugin::loaded() as $pluginName) {
+    if (!empty($amountOfRoutesPerGroup[$pluginName])) {
+        $name = sprintf('%s (%s)', $pluginName, $amountOfRoutesPerGroup[$pluginName]);
+        $pluginNames[$name] = preg_replace('/\W+/', '', strtolower($pluginName));
```
markstory: This could be a call to `Inflector` I think.

templates/element/routes_panel.php
```diff
@@ -19,14 +49,26 @@
...
         $class = '';
-        if ($matchedRoute === $route->template):
-            $class = 'highlighted';
-        elseif ($route->defaults['plugin'] === 'DebugKit'):
-            $class = 'debugkit-route hidden';
+        if (empty($route->defaults['plugin'])) :
+            $class = 'route-entry route-entry--app';
+        else :
+            $class = 'route-entry route-entry--plugin route-entry--plugin-' .
+                preg_replace('/\W+/', '', strtolower($route->defaults['plugin']));
```
markstory: This could be `Inflector` based as well.

webroot/css/toolbar.css
```diff
@@ -512,10 +512,13 @@ pre,
   position: relative;
   top: 2px;
 }
-
 .btn-primary:hover {
   cursor:pointer;
 }
+.toggle-plugin-route-active {
+  background-color: #fff;
+  color: #2a6496;
```
markstory, line 520: I'm not sure I love these colours in this context but I can fiddle with it afterwards.

templates/element/routes_panel.php
LordSimal: I am not able to find a suitable function in the Inflector Utility Class to handle a plugin name like `ADmad/Sequence`

- `Inflector::variable('ADmad/Sequence');` results in `aDmad/sequence` which is not a valid CSS class
- `Inflector::dasherize('ADmad/Sequence');` results in `a-dmad/sequence` which is also not a valid CSS class

Thats why I used this rather simple regex

LordSimal: I would love to have a static cakephp function to just put any string into it and it returns me a valid CSS class.
Currently I have used a custom Helper in my app for that since it is mainly used in the template.

ADmad: Did you try Text::slug()?

LordSimal: this looks good, will push that

LordSimal: To be fair I would have never thought that a `Text::slug()` method could be used to convert a string to a valid CSS class.

PR #853, opened 2021-12-19
fix PHP8.0+ deprecation warning showing Cake/Core/functions.php as source of deprecation

### LINKED ISSUE #845: PHP 8: deprecationWarning() is shown as source of all deprecations
```
CAKE/Core/functions.php:322

QueryExpression::addCase() is deprecated, use case() instead.
```

All deprecations are shown originating in deprecationWarning() even though the file shown in the message is the correct source. 

Can we list the source file instead of functions.php?

LordSimal:
Fixes #845 

With this PR the following deprecation warnings

image

will transform into these deprecation warnings in PHP 8.0+

image

(I had to manually add a `deprecationWarning('Use LocatorAwareTrait::fetchTable() instead.');` into the `ModelAwareTrait::loadModel()` function since it currently has none, only a PHPDoc)

This of course also works inside custom plugins

image

or vendor plugins

image

In all 3 examples I added

```
    $test = $this->loadModel('StaffMembers');

    $test = new QueryExpression();
    $test->addCase([1], 2);
```

as some deprecation test methods.

Basically all I did was to get the current `debug_backtrace()`, iterate over it till I hit a `trigger_error` function, skip one item (it is the `deprecationWarning()` function) and fetch the next items `file` and `line` variables to overwrite the wrong ones coming from the `set_error_handler()` callback.

This is only needed in PHP >= 8.0 as far as I have seen.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -95,13 +95,12 @@ class Plugin extends BasePlugin
             $previousHandler = set_error_handler(
                 function ($code, $message, $file, $line, $context = null) use (&$previousHandler) {
                     if ($code == E_USER_DEPRECATED || $code == E_DEPRECATED) {
-
                         // In PHP 8.0+ the $context variable has been removed from the set_error_handler callback
                         // Therefore we need to fetch the correct file and line string ourselves
                         if (version_compare(PHP_VERSION, '8.0.0') >= 0) {
                             $trace = debug_backtrace();
-                            foreach($trace as $idx => $traceEntry) {
-                                if($traceEntry['function'] !== 'trigger_error') {
+                            foreach ($trace as $idx => $traceEntry) {
+                                if ($traceEntry['function'] !== 'trigger_error') {
                                     continue;
                                 }
                                 $file = $trace[$idx + 2]['file'];
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -104,8 +104,8 @@ class Plugin extends BasePlugin
                                 if($traceEntry['function'] !== 'trigger_error') {
                                     continue;
                                 }
-                                $file = $trace[$idx + 1]['file'];
-                                $line = $trace[$idx + 1]['line'];
+                                $file = $trace[$idx + 2]['file'];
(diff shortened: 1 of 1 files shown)


othercorey:
This will show where the call originated won't include the stack offset, correct?

src/Plugin.php
```diff
@@ -95,6 +95,19 @@ public function setDeprecationHandler($service)
             $previousHandler = set_error_handler(
                 function ($code, $message, $file, $line, $context = null) use (&$previousHandler) {
                     if ($code == E_USER_DEPRECATED || $code == E_DEPRECATED) {
+                        // In PHP 8.0+ the $context variable has been removed from the set_error_handler callback
+                        // Therefore we need to fetch the correct file and line string ourselves
+                        if (version_compare(PHP_VERSION, '8.0.0') >= 0) {
```
ADmad:
```suggestion
                        if (PHP_VERSION_ID >= 80000) {
```

LordSimal:
This shows the place where the deprecated code is being used, not where the deprecated function/class etc. is being defined.

src/Plugin.php
```diff
@@ -95,6 +95,19 @@ public function setDeprecationHandler($service)
...
+                        // In PHP 8.0+ the $context variable has been removed from the set_error_handler callback
+                        // Therefore we need to fetch the correct file and line string ourselves
+                        if (PHP_VERSION_ID >= 80000) {
+                            $trace = debug_backtrace();
+                            foreach ($trace as $idx => $traceEntry) {
+                                if ($traceEntry['function'] !== 'trigger_error') {
+                                    continue;
+                                }
+                                $file = $trace[$idx + 2]['file'];
+                                $line = $trace[$idx + 2]['line'];
```
markstory, line 107: This would be a good place to apply the stack offset defined in the deprecation warning but getting that data here is not straightforward.

Overall this is still an improvement from what we have currently and is now wrong in fewer scenarios. While we can't entirely close out the linked issue we can make incremental progress that is a solution for a large portion of scenarios.

PR #854, opened 2022-01-13
fix deprecation line php8.0+

saeideng:
this lets us to use other value than `1` in second argument of `deprecationWarning()`
also checking `deprecationWarning` instead of `trigger_error` lets to works as before for deprecations that triggered out of  `deprecationWarning()`

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -104,6 +104,7 @@ class Plugin extends BasePlugin
                                     continue;
                                 }
                                 $offset = 1;
+                                // ['args'][1] refers to index of $stackFrame argument in deprecationWarning()
                                 if (isset($traceEntry['args'][1])) {
                                     $offset = $traceEntry['args'][1];
                                 }
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -100,11 +100,15 @@ class Plugin extends BasePlugin
                         if (PHP_VERSION_ID >= 80000) {
                             $trace = debug_backtrace();
                             foreach ($trace as $idx => $traceEntry) {
-                                if ($traceEntry['function'] !== 'trigger_error') {
+                                if ($traceEntry['function'] !== 'deprecationWarning') {
                                     continue;
                                 }
-                                $file = $trace[$idx + 2]['file'];
-                                $line = $trace[$idx + 2]['line'];
+                                $offset = 1;
+                                if (isset($traceEntry['args'][1])) {
+                                    $offset = $traceEntry['args'][1];
+                                }
+                                $file = $trace[$idx + $offset]['file'];
+                                $line = $trace[$idx + $offset]['line'];
                                 break;
                             }
                         }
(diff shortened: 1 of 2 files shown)


src/Plugin.php
```diff
@@ -100,11 +100,15 @@ function ($code, $message, $file, $line, $context = null) use (&$previousHandler
...
                             foreach ($trace as $idx => $traceEntry) {
-                                if ($traceEntry['function'] !== 'trigger_error') {
+                                if ($traceEntry['function'] !== 'deprecationWarning') {
                                     continue;
                                 }
-                                $file = $trace[$idx + 2]['file'];
-                                $line = $trace[$idx + 2]['line'];
+                                $offset = 1;
+                                if (isset($traceEntry['args'][1])) {
+                                    $offset = $traceEntry['args'][1];
```
saeideng, line 109: `1` must be `2` in next version that supports the cake 5

src/Plugin.php
```diff
@@ -100,11 +100,15 @@ function ($code, $message, $file, $line, $context = null) use (&$previousHandler
                         if (PHP_VERSION_ID >= 80000) {
                             $trace = debug_backtrace();
                             foreach ($trace as $idx => $traceEntry) {
-                                if ($traceEntry['function'] !== 'trigger_error') {
+                                if ($traceEntry['function'] !== 'deprecationWarning') {
```
markstory, line 103: Don't we also want to catch `trigger_error` calls? Application or Library code could be using `trigger_error('...', E_USER_DEPRECATED)` and it would be good if debug kit handled that situation.

src/Plugin.php
markstory: Will a test fail so we don't forget?

src/Plugin.php
saeideng: no change in behavior of `trigger_error()` that triggered out of `deprecationWarning()` just like php < 8.0
it means it shows line of calling `trigger_error()`

```php
// src/Controller/AController.php

line 29
line 30 trigger_error("deprecation message", E_USER_DEPRECATED);
line 31

```
//result in both ` php < 8.0` and `> 8.0`
```
App

    APP/Controller/AController.php:30
    deprecation message
```

but before this PR , it was wrong

saeideng: do you say we can add `1` as offset for all `trigger_error()` that triggered out of `deprecationWarning()` ?

saeideng: signature of `deprecationWarning` now (compatible of cake 4) is : 
```
function deprecationWarning(string $message, int $stackFrame = 1)
```
with checking  the `['args'][1]` we want to check used `stackFrame` value 
in direct use of  `trigger_error` , we dont have this info

markstory: ok. I was concerned that we would be capturing the incorrect line number for bare `trigger_error()` calls. It sounds like that is not the case though.

src/Plugin.php
saeideng: added a test

PR #855, opened 2022-01-17
add attributes to request tab

### LINKED ISSUE #714: Add Request attributes to the Request tab
It would be useful to allow exploring the request attributes from the Request panel, specially now we are going to have a lot of useful data related to authentication, like the user identity.

saeideng:
close #714

### DIFF
--- a/templates/element/request_panel.php
+++ b/templates/element/request_panel.php
@@ -19,7 +19,7 @@
 /**
  * @var \DebugKit\View\AjaxView $this
  * @var array $headers
- * @var array $params
+ * @var array $attributes
  * @var array $data
  * @var array $query
  * @var array $cookie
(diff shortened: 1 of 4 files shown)


saeideng:
this will duplicate the `params` key
`params` shows in `Routing Params` section and in this section

templates/element/request_panel.php
```diff
@@ -34,6 +34,9 @@
 <h4><?= __d('debug_kit', 'Routing Params') ?></h4>
 <?= $this->Toolbar->dump($params) ?>
 
+<h4><?= __d('debug_kit', 'Attributes') ?></h4>
+<?= $this->Toolbar->dump($attributes); ?>
```
markstory: Should we remove `Routing Params` instead of having duplicate data?

saeideng: not sure how users feel about `Routing Params` , and they know it is in `attributes` list or no

markstory: That's fair, but if it is the only params list in the panel they will figure it out. We will be leaking some implementation details, but those details are already a public interface. The `getParam()` methods are ease of use methods as the required existence checks to safely use the return of `getAttribute()` are tedious.

saeideng: removed

markstory: We should have an isset check around `$attributes` as history records in the debug kit DB may not have attributes set.

saeideng: `getAttributes()` always have at least 4 items :point_down:  

https://github.com/cakephp/cakephp/blob/41ce80827fd964f6b32e799c837fc3c585a097c1/src/Http/ServerRequest.php#L1674-L1679

saeideng: added

PR #866, opened 2022-05-13
introduce phpstan

LordSimal:


.github/workflows/ci.yml
```diff
@@ -112,11 +112,15 @@ jobs:
...
 
     - name: Run phpcs
       run: vendor/bin/phpcs --report=checkstyle src/ tests/ | cs2pr
 
     - name: Run psalm
       if: success() || failure()
       run: vendor/bin/psalm.phar --output-format=github
+
+    - name: Run phpstan
+      if: success() || failure()
```
ADmad:
```suggestion
      if: always()
```

composer.json
```diff
@@ -32,7 +32,8 @@
     "require-dev": {
         "cakephp/cakephp-codesniffer": "^4.0",
         "cakephp/authorization": "^2.0",
-        "phpunit/phpunit": "~8.5.0 | ^9.3"
+        "phpunit/phpunit": "~8.5.0 | ^9.3",
+        "cakedc/cakephp-phpstan": "^2.0"
```
ADmad: Is this additional dependency really necessary?

src/Panel/DeprecationsPanel.php
```diff
@@ -93,20 +93,9 @@ protected function _prepare()
...
         }
 
-        /** @psalm-suppress RedundantFunctionCall */
-        ksort($return['app']);
-        /** @psalm-suppress RedundantFunctionCall */
-        ksort($return['cake']);
-        /** @psalm-suppress RedundantFunctionCall */
         ksort($return['plugins']);
-        /** @psalm-suppress RedundantFunctionCall */
         ksort($return['vendor']);
```
markstory, line 93: Why are some sorts removed?

composer.json
othercorey: @LordSimal Do you have an update on this? Can we resolve this PR?

src/Panel/DeprecationsPanel.php
ADmad: @markstory These array were lists, not associative arrays, so using key based sorting was redundant.

composer.json
ADmad: I have removed it.

PR #869, opened 2022-06-06
update jQuery to 3.6 and introduce ES6

### LINKED ISSUE #868: Update to latest version of the JQuery 3.6
Update to latest version of the JQuery 3.6 ([Link](https://jquery.com/download/))

LordSimal:
Fixes #868

This updates the jQuery from `v1.11.0` to `v3.6.0`

As you may notice I also introduced ES6 since I asume that we don't support IE11 and already had a prototype class for our toolbar-app.js which is now a proper ES6 class.

The `import` statements above the JS files are browser native dynamic imports, see [HERE](https://caniuse.com/es6-module-dynamic-import) 

I also adjusted the global variable `toolbar` to `debugKitToolbar` to be more specific.

Since we have no real frontend E2E testing here how do we want to test these changes together?


templates/element/history_panel.php
```diff
@@ -73,82 +73,82 @@
...
+                    $('#toolbar').addClass('history-mode');
+                }
 
-            for (var i = 0, len = response.panels.length; i < len; i++) {
-                var panel = response.panels[i];
-                // Offset by two for scroll buttons
-                var button = panelButtons.eq(i + 2);
-                var summary = button.find('.panel-summary');
+                for (var i = 0, len = response.panels.length; i < len; i++) {
+                    var panel = response.panels[i];
```
ADmad:
```suggestion
                    let panel = response.panels[i];
```

templates/element/history_panel.php
```diff
@@ -73,82 +73,82 @@
...
+                    // Offset by two for scroll buttons
+                    var button = panelButtons.eq(i + 2);
```
ADmad:
```suggestion
                    let button = panelButtons.eq(i + 2);
```

templates/element/history_panel.php
```diff
@@ -73,82 +73,82 @@
...
+                    var summary = button.find('.panel-summary');
```
ADmad:
```suggestion
                    let summary = button.find('.panel-summary');
```

ADmad:
It would be nice if all the JS code in the templates was also moved into JS files and the template would just have a single method call.

LordSimal:
Seems like I missed a bunch of `var` in the template files. But yes @ADmad, moving all the JS logic to pure JS files seems a good call and I will do that!

markstory:
> Since we have no real frontend E2E testing here how do we want to test these changes together?

In the past I've manually tested the front-end. Historically browser automation was painful in PHP so I didn't do it. However, in the past few years new libraries like panther have made end to end testing much easier.

webroot/js/debug_kit.js
```diff
@@ -1,25 +1,29 @@
...
-    panelButtons: $('.panel'),
-    panelClose: $('#panel-close'),
-    keyboardScope : $(document),
-    currentRequest: __debugKitId,
-    originalRequest: __debugKitId,
-    baseUrl: __debugKitBaseUrl,
-    webroot: __debugKitWebroot,
-  });
+    let elem = document.getElementById( '__debug_kit_app' );
+    if( elem ) {
```
markstory, line 8: The spacing in these blocks isn't consistent with the rest of the code, can we make it all formatted the same way? Preferably using a linter/formatter so that we can ensure consistency in the future as well.

othercorey, line 8: Is this formatting the preferred standard for js now? `if() {` ? usually it's `if () {`

ADmad: We should stick to as close as possible to what we use for PHP, so `if (..) {`.

markstory: I don't want to get too mired in style nitpicking personally. I can expand the lint task to also check JS and CSS files. I'm thinking of using 'boring' tools like eslint with 'recommended' settings.

LordSimal: Indeed using basic eslint was also my plan.
But first I have to do the base work of actually refactoring the current jQuery mess into ES6 modules.
After the functional part is over we can then argue about what JS code styling we use...

LordSimal:
As discussed in the core chat I will refactor the whole frontend part of this project and try to implement eslint as well for the new ES6 JS Modules

Issue #871, opened 2022-06-07
Excessive Memory usage after upgrading to DebugKit 4.8.0 (from 4.7.1) in Testsuite


swiffer:
Memory usage / Execution Time in DebugKit is crazy after updating to DebugKit 4.8.0.
Anyone else noticing this issue? It's a test for one controller only - running the full TestSuite results in out of memory errors

with DebugKit 4.7.1
```bash
> composer test -- --filter PagesControllerTest
> phpunit --colors=always --filter PagesControllerTest
PHPUnit 9.5.20 #StandWithUkraine

......                                                              6 / 6 (100%)

Time: 00:00.395, Memory: 18.00 MB

OK (6 tests, 15 assertions)
```

with DebugKit 4.8.0
```bash
> composer test -- --filter PagesControllerTest
> phpunit --colors=always --filter PagesControllerTest
PHPUnit 9.5.20 #StandWithUkraine

......                                                              6 / 6 (100%)

Time: 00:01.587, Memory: 414.00 MB

OK (6 tests, 15 assertions)
```

LordSimal:
Sorry, this is now fixed in the new 4.8.1 version. 
Please update your Debug Kit and this problem should be gone.

swiffer:
Thanks addressing this issue so quickly!

PR #873, opened 2022-06-08
add a test to check that DebugkitMiddleware is not loaded in phpunit run

LordSimal:
Refs: #871

### DIFF
--- a/tests/TestCase/PluginTest.php
+++ b/tests/TestCase/PluginTest.php
@@ -17,9 +17,11 @@ namespace DebugKit\Test\TestCase;
 use Cake\Error\PhpError;
 use Cake\Event\Event;
 use Cake\Event\EventManager;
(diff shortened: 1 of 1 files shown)


tests/TestCase/PluginTest.php
```diff
@@ -70,12 +72,28 @@ public function testSetDeprecationHandler()
...
         $this->assertEquals($parsed['file'], 'src/Plugin.php');
         $this->assertEquals($parsed['line'], 51);
     }
+
+    /**
+     * Make sure the DebugkitMiddleware is not loaded while running PHPUnit
+     *
+     * @return void
+     */
+    public function test_middleware_not_loaded_in_tests()
```
othercorey: This should follow the standard naming convention.

LordSimal: Sorry, still used to snake style from my work 🙈

PR #878, opened 2022-06-18
4.next - New JS & CSS

LordSimal:
Well, here we go 😄 
This whole story started with #868 and the need to update jQuery from 1 to 3. I tried to do that in #869 but quickly ran into major refactoring problems. Thats why this PR exists now.

# General Structure

It took me some time to realize how the whole iframe initialization worked but let me just sum it up here:

## App HTML

The `DebugKitMiddleware` is responsible to inject a script tag into your app with the `webroot/js/inject-iframe.js` src and other data attributes.

This JS file injects an iframe at the end of the `body` with the src `/debug-kit/toolbar/<request-id>` and triggers further functionality inside the iFrame.

## iFrame HTML
* `templates/Requests/view.php` includes both `jQuery` and the `webroot/js/main.js` JS file
* This file initializes the Toolbar and loads all other ES6 modules to setup the event listeners for each panel
* `webroot/js/modules/Toolbar.js::loadPanel` is responsible to load a specific panel

# What did I do?

## New CSS class naming schema
The first major part was to go through all the templates and introduce a more [BEM](http://getbem.com/) oriented class structure combined with component based CSS.

The TLDR for BEM (who hasn't heard of it) is `.block__element--modifier`. In our example each panel is its own block which has elements and some of these elements (or the block itself) can be modified (e.g. `.o-list--deprecation-list`). 

Also I introduce the concept of `components` and `objects` which prefix each class with either `c-` or `o-`.
The main difference between those 2 is the fact, that `objects` are more generic (like inputs, buttons, lists etc.) and components are more specific.

## SCSS or pure CSS?
Currently I have set it up to generate a single `style.css` from multiple `scss` files present in the `css-source` folder via [Laravel Mix](https://laravel-mix.com/).
If you want to adjust CSS you can start the watcher via `npm run start` or build a minified version via `npm run build`.

The main question here now is: Do we want to switch to the new SCSS generated setup or stay at the goold old plain CSS path? I can of course just generate the CSS now and commit it that way so we need to adjust the `style.css` file directly for ruther changes. But if so then I wouldn't keep the `css-source` folder present to not have 2 different CSS sources present in the repo. 

The only disadvantage I see we have for the generated CSS setup is the fact, that we have to regenerate the `style.css` everytime a change is done in the source files before we tag a new release.

## JS Changes
The first major JS change was the introduction of ES6 Modules to split up the JS functionality into separate files so its easier to handle. So basically every `import` you can see at the start of `webroot/js/main.js` is a JS module and will be loaded as a separate request automatically when JS is initialized in the iframe.

Talking about the iFrame: The whole iFrame injection logic has been cleaned up in the `webroot/js/inject-iframe.js` as well.

## Remove jQuery?
Well, I tried to remove jQuery but the fact that we use `.ajax()` as well as the `:contains` selector lead me to the current situation, that I left jQuery in there, just updated to 3.6 as desired.

## Introduce Code Style for JS
I personally haven't used any Code Style Tools for JS code but my colleagues (and basically the first thing you find on google) always refers to the [Airbnb JS Code Style](https://airbnb.io/javascript/)

Therefore I setup eslint with the `airbnb-base` ruleset but I still had to adjust some rules to not have to adjust code even further. You can check JS style code via `npm run cs-check` or automatically let it fix stuff with `npm run cs-fix`

## Moved mostly all JS functionality away from panel template files
Another desired change was the fact, that currently we have JS functionality present for each Panel inside the template file for that specific panel. 
This also changed in the fact, that each Panel has its own JS file in `webroot/js/modules/Panels`.
The initialization happens via a 
```
document.dispatchEvent(new CustomEvent('initPanel', { detail: `panel${panelType}` }));
```
inside the `webroot/js/modules/Toolbar.js::loadPanel()` function.

With that I tried to simplify the `Toolbar` class which previously had most of that logic in there.

The `mostly` part relates to the `formatChanged()` JS functionality present in `templates/MailPreview/email.php` which I didn't change due to the fact, that the Email Preview template doesn't contain.

## How can I test this myself?
The easiest way for me was to open up a CakePHP project with (partially) functionality present which can be used to test the DebugKit (like sending emails) and add a symlink from `vendor/cakephp/debug_kit` to wherever you have this repo checked out.

So for me
```
ln -s /Users/kevinpfeifer/Documents/CakePHP/debug_kit/ vendor/cakephp/debug_kit
```
resulted in
```
kevinpfeifer@Kevins-MBP [13:15:48] [~/Documents/Sunlime/alfred/vendor/cakephp] [master *]
-> % ls -al debug_kit               
lrwxr-xr-x  1 kevinpfeifer  staff  48 Jun 11 20:54 debug_kit -> /Users/kevinpfeifer/Documents/CakePHP/debug_kit/
```
and I had easy access to test and change code directly (expecting that destination folder has the `4.next-new-js-css` branch checked out 😉 )

## Removed deprecations
Other small changes relate to 
* Replacing `App::path` with `App::classPath` in `src/Controller/MailPreviewController.php`
* Replace `ModelAwareTrait` with `LocatorAwareTrait` in `src/Mailer/MailPreview.php`

## Improved loading animation

Still have to debug the stutter while animating to the top when the panel has much data inside it


othercorey:
> The only disadvantage I see we have for the generated CSS setup is the fact, that we have to regenerate the style.css everytime a change is done in the source files before we tag a new release.

I think that would have to be for each PR/commit. I don't think we can rely on generating files for each tag.

markstory:
A few high level comments before I dig into the diffs.

> It took me some time to realize how the whole iframe initialization worked but let me just sum it up here

I admit how this works is not the most approachable solution but we needed to use an iframe to ensure that the toolbar had a reliable CSS and DOM environment to exist in.

> Do we want to switch to the new SCSS generated setup or stay at the goold old plain CSS path? I can of course just generate the CSS now and commit it that way so we need to adjust the style.css file directly for further changes. 

My personal bias would be to keep only the CSS. While SCSS is nice to author the additional tooling complexity and needing to remember to update the repo with the built assets makes feel that we should work with what we get from CSS. With that said the styles don't need to use a monolithic file. We can use `@import()` to load 'modules' of CSS.

> Therefore I setup eslint with the airbnb-base ruleset but I still had to adjust some rules to not have to adjust code even further. 

This is a great set of conventions to use. We can remove the overrides in the future by doing a pull request that is just reformatting.

>  This also changed in the fact, that each Panel has its own JS file in webroot/js/modules/Panels.

Neat solution, I like it. :clap:

.jshintrc
```diff
@@ -1,84 +1,16 @@
...
-
-    // Legacy
-    "nomen"         : false,    // true: Prohibit dangling `_` in variables
-    "onevar"        : false,    // true: Allow only one `var` statement per function
-    "passfail"      : false,    // true: Stop on first error
-    "white"         : true,     // true: Check against strict whitespace and indentation rules
-
-    // Custom Globals
-    "predef"        : [ ]       // additional predefined global variables
+    "browser": true,
```
markstory: If we're using eslint, do we need jshint as well?

css-source/components/_debug-table.scss
```diff
@@ -0,0 +1,28 @@
...
+    word-break: break-word; /* Not standard for webkit */
+}
+
+.c-debug-table th.left,
+.c-debug-table td.left {
+    text-align: left;
+}
+
+.c-debug-table .duplicate-route td {
+    background: #ffea96;
```
markstory: Should we be using CSS custom properties for colors?

package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "debug_kit",
+    "version": "1.0.0",
+    "description": "CakePHP Debug Kit",
+    "author": "Kevin Pfeifer",
```
markstory: Can we change this to `CakePHP Team`?

templates/MailPreview/email.php
```diff
@@ -26,8 +26,8 @@
 
 <script>
     function formatChanged(form) {
-        var part_name = form.options[form.selectedIndex].value
-        var iframe = document.getElementsByName('messageBody')[0];
+        let part_name = form.options[form.selectedIndex].value
+        let iframe = document.getElementsByName('messageBody')[0];
         iframe.contentWindow.location.replace('?part=' + part_name + '&plugin=<?= h($plugin); ?>');
```
markstory, line 31: We might be able to move this into an event handler that reads from global variables, or variables that are set into the toolbar script as script tag attributes. :shrug:

webroot/js/modules/Panels/CachePanel.js
```diff
@@ -0,0 +1,35 @@
...
+          addMessage(errorThrown);
+        },
+      });
+      e.preventDefault();
+    });
+  };
+
+  return {
+    init,
+  };
```
markstory: Should this use the `onEvent` pattern you used in Mail and History panels?

webroot/js/modules/Panels/RoutesPanel.js
```diff
@@ -0,0 +1,20 @@
...
+        $(`.c-routes-panel__route-entry${plugin}`).removeClass('is-hidden');
+      } else {
+        $this.addClass('is-active');
+        $(`.c-routes-panel__route-entry${plugin}`).addClass('is-hidden');
+      }
+    });
+  };
+
+  return {
+    init,
```
markstory: Should this use the `onEvent` pattern?

webroot/js/modules/Toolbar.js
```diff
@@ -0,0 +1,257 @@
...
+
+      // This initializes the panel specific JS logic (if there is any)
+      document.dispatchEvent(new CustomEvent('initPanel', { detail: `panel${panelType}` }));
+      that.bindNeatArray();
+      that.bindDebugBlock();
+    });
+  }
+
+  // ???????
+  bindNeatArray() {
```
markstory: This can be removed.

css-source/components/_debug-table.scss
LordSimal: I will introduce CSS Variables when removing the SCSS code

package.json
LordSimal: Sure, was just auto generated 😄

.jshintrc
LordSimal: I first thought the .jshintrc was needed for IDE purposes but there is also eslint support for at least PHPStorm.
So yes, we can drop it.

webroot/js/modules/Toolbar.js
LordSimal: So can I remove the `ToolbarHelper::makeNeatArray()` method as well?

webroot/js/modules/Panels/CachePanel.js
LordSimal: I usually prefer to use the `$(document).on('click', '.js-some-class', function(){}` because this only has to be executed once and it works even with content that gets added via AJAX. But if this gets excecuted multiple times then we have mutliple event listeners on the same element which in this case would trigger multiple ajax requests to clear the cache.

But since we now have a pretty clear initialization event for each panel this is not really necessary and therefore I can change that back to `$('.js-some-class').on('click', function(){}` so the event listeneres get re-added everytime a panel is reloaded.

LordSimal:
I am going to ignore the SticklerCI complaints here 😅  (see above commits where it introduces errors itself)

othercorey:
> I am going to ignore the SticklerCI complaints here 😅 (see above commits where it introduces errors itself)

Just disable fixing/commits in the stickler config file.

webroot/js/modules/Toolbar.js
markstory: Yes. All of that has been replaced with newer features in `Debugger`.

templates/MailPreview/email.php
LordSimal: Do we really need to introduce another JS file just for those 4 Lines of JS?
This is only needed on the https://myurl/debug-kit/mail-preview/preview/MyMailPreview/someAction?plugin= template which only has a very minimal amount of CSS and JS.

markstory (approved):
I did some local testing and pushed up a couple commits to improve styling on the routes and packages panel.

Nice work on modernizing the CSS and JS for this project without pulling in more complexity :clap:

webroot/css/style.css
```diff
@@ -283,6 +283,10 @@ strong {
 .c-debug-table .highlighted td {
     background: var(--highlighted);
 }
+.c-debug-table th.c-left-align,
+.c-debug-table td.c-left-align {
```
LordSimal: I would rather create a `.u-text-left` at the end of the style.css and use it here, therefore it can  be re-used in other components as well

templates/MailPreview/email.php
markstory: That's fair Lets leave this here then.

PR #879, opened 2022-06-21
Update for Cake 5.

ADmad:
Two things remaining:

1. Add back authorization plugin to "require-dev" when it has a Cake 5 compatible branch.
2. Update `LazyTableTrait` for schema creation. `TestFixture::create()` no longer exists.


composer.json
```diff
@@ -23,16 +23,15 @@
         "source": "https://github.com/cakephp/debug_kit"
     },
     "require": {
-        "php": ">=7.4",
-        "cakephp/cakephp": "^4.4.0",
+        "php": ">=8.1",
```
othercorey: Are we going to drop this for plugins?

othercorey:
I don't think removing .stickler.yml disables the check. We can just disable commits.

ADmad:
> I don't think removing .stickler.yml disables the check. We can just disable commits.

We could just remove the Stickler app from the organization. It would be a pain having to update it's config for each `cake5` branch for each repo.

ADmad:
I have created #880 to track pending changes. This can be merged.

src/Cache/Engine/DebugEngine.php
```diff
@@ -296,21 +291,21 @@ public function groups(): array
...
-     * @param string|array $key The key to set, or a complete array of configs.
+     * @param array|string $key The key to set, or a complete array of configs.
      * @param mixed|null $value The value to set.
      * @param bool $merge Whether to recursively merge or overwrite existing config, defaults to true.
      * @return $this
-     * @throws \Cake\Core\Exception\Exception When trying to set a key that is invalid.
+     * @throws \Cake\Core\Exception\CakeException When trying to set a key that is invalid.
      */
-    public function setConfig($key, $value = null, $merge = true)
+    public function setConfig(string|array $key, mixed $value = null, bool $merge = true)
```
othercorey: Unfortunately, the sniff that converts annotations to type declarations seems to run first so we get out-of-order type declarations. I just searched and replaces the 2 most common: string|array and array|ArrayAccess.

src/Cache/Engine/DebugEngine.php
```diff
@@ -65,13 +57,18 @@ class DebugEngine extends CacheEngine
     /**
      * Constructor
      *
-     * @param array<string, mixed>|\Cake\Cache\CacheEngine $config Config data or the proxied adapter.
+     * @param \Cake\Cache\CacheEngine|array<string, mixed> $config Config data or the proxied adapter.
      * @param string $name The name of the proxied cache engine.
      * @param \Psr\Log\LoggerInterface $logger Logger for collecting cache operation logs.
      */
-    public function __construct($config, string $name, LoggerInterface $logger)
+    public function __construct(array|CacheEngine $config, string $name, LoggerInterface $logger)
```
othercorey: This should be `CacheEngine|array`

src/Log/Engine/DebugKitLog.php
```diff
@@ -26,17 +27,17 @@ class DebugKitLog extends BaseLog
...
      * Captures log messages in memory
      *
      * @param mixed $level The type of message being logged.
-     * @param string $message The message being logged.
+     * @param string|\Stringable $message The message being logged.
      * @param array $context Additional context data
      * @return void
      */
-    public function log($level, $message, array $context = [])
+    public function log(mixed $level, string|Stringable $message, array $context = []): void
```
othercorey: `Stringable|string`

src/ToolbarService.php
```diff
@@ -225,9 +227,9 @@ public function initializePanels()
      *
      * @param \Cake\Http\ServerRequest $request The request
      * @param \Psr\Http\Message\ResponseInterface $response The response
-     * @return false|\DebugKit\Model\Entity\Request Saved request data.
+     * @return \DebugKit\Model\Entity\Request|false Saved request data.
      */
-    public function saveData(ServerRequest $request, ResponseInterface $response)
+    public function saveData(ServerRequest $request, ResponseInterface $response): false|Request
```
othercorey: `Request|false`

tests/TestCase/Panel/VariablesPanelTest.php
```diff
@@ -62,21 +63,21 @@ public function testShutdown()
...
-        $unserializableDebugInfo->expects($this->any())->method('__debugInfo')->willReturn([
-            'unserializable' => $unserializable,
-        ]);
+        // $unserializableDebugInfo = $this
+        //     ->getMockBuilder('\stdClass')
+        //     ->addMethods(['__debugInfo'])
+        //     ->getMock();
+        // $unserializableDebugInfo->expects($this->any())->method('__debugInfo')->willReturn([
+        //     'unserializable' => $unserializable,
+        // ]);
```
othercorey: Why is this commented out?

ADmad: It's no longer possible to get the PDO instance from the driver. I couldn't figure out an alternative at the time.

src/Cache/Engine/DebugEngine.php
ADmad: It's going to be a PITA if this can't be auto fixed.

othercorey: When we updated cakephp, I ran the 5.x phpcs without native types on 4.next which fixed only the annotations. Then in 5.x, I ran with native types with the already fixed annotations.

If phpcs gives me a list of native types, then I could sort them, but if I have to handle line breaks, that's not going to work.

othercorey: Also, you don't have to fix these if we don't consider it an issue.

ADmad: I fixed these but unless we can auto fix it we shouldn't enforce such ordering.

ADmad:
I can't figure out why the tables are missing on GA, causing the test failures. If I run the tests locally without setting `DB_URL` in environment causing [tests/bootstrap](https://github.com/cakephp/debug_kit/blob/8068b345e1d503e0e9362bd14e7b430f04bc6560/tests/bootstrap.php#L90) to use Sqlite with a db file then there are no errors.

Going to merge this anyway as it would allow working on other plugins which have debugkit as a dependency, for e.g. bake.

PR #881, opened 2022-06-29
Update database schema generation to not rely on fixtures

markstory:
Remove the need to load/use fixture code in debug kit. Instead we can
use the same schema data format as the internal test schema loader and
duplicate a small passage to ensure we have stability.


ADmad:
I took care of the CS failures and added back `cakephp/authorization`.

PR #884, opened 2022-08-12
Add doc section to explain how to use toolbar with an api only project

LordSimal:
As mentioned by @cnizzardini its currently not clear how to use the DebugKit Toolbar with an API only project due to the fact, that we only mention the automatically present toolbar in the frontend.

This new section inside the doc should clarify that using the DebugKit is also possible for API only projects, just a little bit more tedious because one has to manually call the URL from the iframe.

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -291,6 +291,20 @@ to include the panel::
 The above would load all the default panels as well as the ``AppPanel``, and
 ``MyCustomPanel`` panel from ``MyPlugin``.
 
+Accessing Toolbar without a frontend
+====================================
+
+If you have an application which only provides an API (and therefore no frontend)
+the usual way of accessing the toolbar can't be used.
+
+Instead you have to call http://localhost/debug-kit/toolbar/``<debugkit-id>``
+
+The ``<debugkit-id>`` can be found inside the HTTP headers of your API response. It should look something like that::
+
+    X-DEBUGKIT-ID: 5ef39604-ad5d-4ca4-85d8-8595e52373bb
+
+So you would have to call http://localhost/debug-kit/toolbar/5ef39604-ad5d-4ca4-85d8-8595e52373bb
+
 Helper Functions
 ================
 


markstory:
Should we have a standalone debugkit view for API applications? I'm imagining something that has a list of requests in a sidebar and the main 'toolbar' as tabs across the top/bottom of the page?

LordSimal:
I think you mean something like a generic http://localhost/debug-kit/toolbar/ view which can be called no matter what type of project one is running.
And inside that all saved requests (currently the content of the "History" tab) are being shown inside a e.g. sidebar so one can easily choose which request they want to debug. Seems like a good idea to me.

cnizzardini:
Thanks. I believe this does something similar to what is being discussed: https://github.com/itsgoingd/clockwork

raul338:
#834 was for this specific use case.

Go to http://localhost/debug-kit/dashboard and refresh the requests panel, you can see the api requests in there

LordSimal:
@raul338 I didn't know that this was present but I also couldn't find any mention of it in the documentation.
Can you maybe add a section explaining the dashboard and its connection to the history panel comming from an API only aspect?

cnizzardini:
@LordSimal I finally had a chance to look at this again. It seems if I click on the "History" tab I can then select a specific request and then view the SQL, variables etc. for that request.

The part which seems broken (or maybe just bad UX) is that the History panel says "0 XHR" even after refreshing. 

image

If this is a bug I can create an issue for it.

LordSimal:
@cnizzardini Its not really a bug because the `0 xhr` means that your curent browser session hasn't executed any XMLHttpRequest requests (yet).

This `0 xhr` only updates if your current active browser session performs an XMLHttpRequest while the debug_kit toolbar is open/active because we are overwriting the `window.XMLHttpRequest.prototype.send` method when the iFrame is injected.

Therefore we are not saving XMLHttpRequest requests inside the server and can't "refresh" them when switching base request entities.

PR #890, opened 2022-09-10
4.next - Switch to `doctrine/sql-formatter`.

ndm2:
There's no updates anymore for `jdorn/sql-formatter`, PRs aren't getting merged, it's dead. `doctrine/sql-formatter` is a fork with additional functionality such as CLI highlighting, also I've added support for common table expressions and window functions to it.

I've opted to test for Sqlite only for the output tests, because accounting for the different quote characters is just very annoying, and it's not really overly important.

# HTML before:

html-before

# HTML after:

html-after

# CLI after:

cli

### DIFF
--- a/src/DebugSql.php
+++ b/src/DebugSql.php
@@ -18,7 +18,10 @@ namespace DebugKit;
 use Cake\Core\Configure;
 use Cake\Database\Query;
 use Cake\Error\Debugger;
-use SqlFormatter;
+use Doctrine\SqlFormatter\CliHighlighter;
+use Doctrine\SqlFormatter\HtmlHighlighter;
+use Doctrine\SqlFormatter\NullHighlighter;
+use Doctrine\SqlFormatter\SqlFormatter;
 
 /**
  * Contains methods for dumping well formatted SQL queries.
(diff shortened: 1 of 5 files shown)


ndm2:
Not sure what's up with Codecov, unless I'm blind the lines are covered 🤷

ndm2:
Ugh, OK, coverage runs on MySQL, let me get that fixed 🥲

src/DebugSql.php
```diff
@@ -93,10 +96,8 @@ public static function sql(Query $query, $showValues = true, $showHtml = null, $
         }
 
         $template = self::$templateHtml;
-        $sqlHighlight = true;
-        if ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') || $showHtml === false) {
```
othercorey, line 97: This could probably be a global function in cakephp.

src/DebugSql.php
```diff
@@ -143,6 +161,16 @@ public static function sqld(Query $query, $showValues = true, $showHtml = null,
...
 
+    /**
+     * Checks whether the current environment is CLI based.
+     *
+     * @return bool
+     */
+    protected static function isCli()
+    {
+        return PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg';
+    }
```
othercorey, line 172: This could probably be a global function in cakephp.

othercorey:
Looks good!

src/DebugSql.php
ndm2: Maybe. The core seems to differentiate in some areas between `cli` only, and `cli || phpdbg`.

ndm2:
OK, now it tests for fragments only to figure whether the expected type of output is being generated, eg HTML, or ANSI codes, or plain text. It assumes that the library is working as expected.

othercorey:
> OK, now it tests for fragments only to figure whether the expected type of output is being generated, eg HTML, or ANSI codes, or plain text. It assumes that the library is working as expected.

This is probably better long-term. If they change formatting slightly, we don't want our unit tests to fail.

PR #895, opened 2022-11-03
Remove translations

garas:
- Files were in wrong locations
- File contents untouched since 2009
- Only 8 of 30 translated strings in `fra`, `nld` and `pt_BR` are still used until now, other languages are empty
- There are 124 translatable strings now, only 8 strings translated
- It seems nobody cared about incomplete or non-existing translations

Should I replace `__d('debug_kit', 'Text')` to plain strings too?


ADmad:
> Should I replace __d('debug_kit', 'Text') to plain strings too?

I am in favor of doing so. We don't have the manpower or interest to maintain proper translations for a debugging tool.

othercorey:
Seems ok, but doesn't affect me. We should track changes in releases if we were trying to support this for translators.

markstory:
Makes sense to me too. We've never had anyone interested in maintaining translations so they aren't benefiting anyone right now.

PR #897, opened 2022-11-11
Fix Reload button in History panel

garas:
After JS+CSS refactor, button stopped working.

It got wrong `this` with short syntax.

I'm bit outdated on JS, so I don't know difference between syntaxes. I thought its just shorter syntax.


LordSimal:
Thanks for catching that bug.
Yes, the new JS short syntax `() => {}` does **not** have a `this` context, I missed that one when I did the refactor.
This indeed fixes that bug because a "normal" `function() {}` does have a `this` context.

Issue #857, opened 2022-03-15
[feature request] Change iframe size when debugKit iframe shows an error


hacheraw:
![image](https://user-images.githubusercontent.com/10481749/158342336-2f8c9e23-f955-4a8d-af76-94de10911449.png)

In some cases, when there's an application error (usually error 5xx or routes one) the DebugKit iframe looks like the above picture.

Would it be possible to add a css class to the iframe if an error template is shown inside? Or just change it's size.

**Wanted behavior:**
iFrame is 40x40px by default. When an error is thrown, iFrame should be, maybe, 40% by 40% so we can see the error.

![image](https://user-images.githubusercontent.com/10481749/158342237-15fef1cd-dfad-43a4-bc37-b2621b3d26ea.png)

LordSimal:
I'll try to implement that now.

LordSimal:
The mentioned PR above only catches errors which happen when a panel like `Requests` or `Routes` is loaded.

But can you explain a scenario where the initial DebugKit Toolbar is not loading / throwing an error? This would need be handled seperately I would say.

hacheraw:
Sorry, I don't know if I'm understanding your question but I can't remember how to reproduce it in different scenarios.

For instance, that screenshot was from a day I was trying to implement a custom Middleware. Adding a bug on a custom Middleware could be a way to reproduce this.

It's been a couple of months since I don't touch CakePHP, so I'm a bit rusty...

hacheraw:
Now looking at your gif, yeah, I understand what you mean.

I'll try to reproduce it before Tuesday.

LordSimal:
That error would have to be somewhat `debug_kit` specific because middlewares are run everywhere (not only inside debug_kit)

Therefore if you have an error in your middleware it would be shown in the "normal" window, not the DebugKit window.

But if you find a way how to reproduce that it would be awesome 👍🏻

hacheraw:
If I recall correctly, I was playing with redirects when the url has /es /de, etc...
And someway it made the app work but /debug-kit route was not loading.

So try making /debug-kit route not loading or wait for me to try to reproduce.

Sorry, I'm not at home/work now

LordSimal:
I'll try that, thanks 👍🏻 
And no stress, thanks for the response after quite some time ^^

hacheraw:
**Use this!**
The middleware will redirect debug_kit request to a non-existent Controller.
[...]

PR #898, opened 2022-11-11
4.x: show debug_kit iframe error nicely

LordSimal:
Refs https://github.com/cakephp/debug_kit/issues/857

This adds a nice catch when a panel can't be loaded due to some error on the server side.

This behavior can be reproduced if you add a syntax error inside the `vendor/cakephp/debug_kit/src/Controller/PanelsController.php` view method.

### DIFF
--- a/webroot/js/modules/Toolbar.js
+++ b/webroot/js/modules/Toolbar.js
@@ -152,7 +152,14 @@ export default class Toolbar {
       // This initializes the panel specific JS logic (if there is any)
       document.dispatchEvent(new CustomEvent('initPanel', { detail: `panel${panelType}` }));
       that.bindDebugBlock();
-    });
+    })
+      .fail((response) => {
+        clearTimeout(timer);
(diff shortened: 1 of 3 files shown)


webroot/js/modules/Toolbar.js
```diff
@@ -152,7 +152,14 @@ export default class Toolbar {
       // This initializes the panel specific JS logic (if there is any)
       document.dispatchEvent(new CustomEvent('initPanel', { detail: `panel${panelType}` }));
       that.bindDebugBlock();
-    });
+    })
+      .fail(() => {
+        clearTimeout(timer);
+        $(`<iframe src="${url}" width="100%" height="100%"></iframe>`).appendTo(contentArea);
```
markstory: Why use another iframe instead of updating the content of `contentArea`?

LordSimal: I first wasn't able to get that to work for some reason but you are right, we don't need another iframe here.

PR #901, opened 2022-12-20
5.x: fix tests

LordSimal:
No idea why this changed but seems to me like we adjusted some offset in the Debugger.

Also there are errors related to `$xml->xpath($path);` where I have no idea as well because I never worked with `xpath()`.

image

Will have to look into that...


LordSimal:
Found it. We changed the class names for the generated debug HTML elements

PR #906, opened 2023-01-14
5.x: reimplement #574 in Cake5

### LINKED ISSUE #573: Show the originating PHP file in the SQL log

LordSimal:
Reimplements #574
Closes #573

I really liked this idea and reimplemented it for the Cake5 branch.
Kudos definitely go to @thinkingmedia for implementing it in the first place but for some reason it didn't get finished.

The only thing that we could maybe adjust is the position of the SQL comment. Does someone else maybe have a better idea of placing it in the query?

Currently it would look like this:
image

### DIFF
--- a/src/Model/Table/PanelsTable.php
+++ b/src/Model/Table/PanelsTable.php
@@ -34,6 +34,7 @@ use RuntimeException;
 class PanelsTable extends Table
 {
     use LazyTableTrait;
+    use SqlTraceTrait;
 
     /**
      * initialize method
--- a/src/Model/Table/RequestsTable.php
+++ b/src/Model/Table/RequestsTable.php
@@ -36,6 +36,7 @@ use PDOException;
 class RequestsTable extends Table
 {
     use LazyTableTrait;
+    use SqlTraceTrait;
 
     /**
      * initialize method
--- a/src/Model/Table/SqlTraceTrait.php
+++ b/src/Model/Table/SqlTraceTrait.php
@@ -1,0 +1,95 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         5.0.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace DebugKit\Model\Table;
+
+use Cake\Core\Configure;
+use Cake\Error\Debugger;
+use Cake\ORM\Query\DeleteQuery;
+use Cake\ORM\Query\SelectQuery;
+use Cake\ORM\Query\UpdateQuery;
+
+/**
+ * Add this trait to your Table class to append the file reference of where a Query object was created.
+ *
+ * @mixin \Cake\ORM\Table
+ */
+trait SqlTraceTrait
+{
+    /**
+     * Overwrite parent table method to inject SQL comment
+     */
+    public function selectQuery(): SelectQuery
+    {
+        return $this->fileStamp(parent::selectQuery());
+    }
+
+    /**
+     * Overwrite parent table method to inject SQL comment
+     */
+    public function updateQuery(): UpdateQuery
+    {
+        return $this->fileStamp(parent::updateQuery());
+    }
+
+    /**
+     * Overwrite parent table method to inject SQL comment
+     */
+    public function deleteQuery(): DeleteQuery
+    {
+        return $this->fileStamp(parent::deleteQuery());
+    }
+
+    /**
+     * Applies a comment to a query about which file created it.
+     *
+     * @template T of \Cake\ORM\Query\SelectQuery|\Cake\ORM\Query\UpdateQuery|\Cake\ORM\Query\DeleteQuery
+     * @param \Cake\ORM\Query\SelectQuery|\Cake\ORM\Query\UpdateQuery|\Cake\ORM\Query\DeleteQuery $query The Query to insert a comment into.
+     * @psalm-param T $query
+     * @param int $start How many entries in the stack trace to skip.
+     * @param bool $debugOnly False to always stamp queries with a comment.
+     * @return \Cake\ORM\Query\SelectQuery|\Cake\ORM\Query\UpdateQuery|\Cake\ORM\Query\DeleteQuery
+     * @psalm-return T
+     */
+    protected function fileStamp(
+        SelectQuery|UpdateQuery|DeleteQuery $query,
+        int $start = 1,
+        bool $debugOnly = true
+    ): SelectQuery|UpdateQuery|DeleteQuery {
+        if (!Configure::read('debug') && $debugOnly === true) {
+            return $query;
+        }
+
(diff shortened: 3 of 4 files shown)


ADmad:
Can we have the comments above `SELECT` instead of below it?

LordSimal:
Well I would love to not only adjust it inside the debug kit SQL panel output but also when someone calls `$query->sql()`.
But that would require us to adjust the QueryCompiler in the core as far as I understand. But I haven't looked that deep into that one yet.

markstory:
Is the idea that developers will use this trait in their tables temporarily when they want to add tracing to the queries?

LordSimal:
> Is the idea that developers will use this trait in their tables temporarily when they want to add tracing to the queries?

currently yes.

LordSimal:
This would of course kind of restrict users from adding that trait to e.g. table classes inside vendor plugins unless they modify the file inside vendor. I still like the idea of having only specific models which get this kind of additional info being injected.

LordSimal:
With the new `->comment()` method provided by the core this now looks like this:
image

mehov:
> This would of course kind of restrict users from adding that trait to e.g. table classes inside vendor plugins unless they modify the file inside vendor.

On my local development server, I put the trait straight into `vendor/cakephp/cakephp/src/ORM/Table.php` hoping it traces all queries not only from within my host application, but also from all plugins that it uses, because that's the ultimate table class that everyone ends up extending. But it didn't work.

I tested by going to `https://example.local:443/profile` [hosted](https://github.com/CakeDC/users/blob/9f9f49bb8a0f0302ddfab46195367b1f73d698da/src/Controller/Traits/ProfileTrait.php#L33) by [`cakedc/users`](https://github.com/CakeDC/users) where it makes a query to the users table.

Here's what I added to `vendor/cakephp/cakephp/src/ORM/Table.php` (with context):

```
//…

class Table implements RepositoryInterface, EventListenerInterface, EventDispatcherInterface, ValidatorAwareInterface
{
    /**
     * @use \Cake\Event\EventDispatcherTrait<\Cake\ORM\Table>
     */
    use EventDispatcherTrait;
    use RulesAwareTrait;
    use ValidatorAwareTrait;

// start
    //book.cakephp.org/debugkit/5/en/index.html#tracing-query-execution
    use \DebugKit\Model\Table\SqlTraceTrait;
// stop

    /**
     * Name of default validation set.
     *
     * @var string
     */
    public const DEFAULT_VALIDATOR = 'default';

    // … etc …
```

When I put a wrong namespace into that `use` statement I added, it errors out, which proves that `\CakeDC\Users\Model\Table\UsersTable` [does indeed extend](https://github.com/CakeDC/users/blob/9f9f49bb8a0f0302ddfab46195367b1f73d698da/src/Model/Table/UsersTable.php#L40) `\Cake\ORM\Table`. But the tracing is not applied still.

mehov:
I figured out why as soon as I submitted the comment. The trait hijacks some query-related methods from the parent table class. But because I'm applying the trait to `\Cake\ORM\Table` itself, there is no parent, as `\Cake\ORM\Table` goes on to define these methods, overwriting what's in the trait.

So I have to either now hijack `\Cake\ORM\Table`, or have all my plugins - at least those I control - use some kind of shared `App\Model\Table\AppTable`.

**UPD** guess what I went with :D

Posting here in case someone else is looking for this.

```
cp vendor/cakephp/cakephp/src/ORM/Table.php vendor/cakephp/cakephp/src/ORM/TableFromCake.php
```

Rename the class in `TableFromCake.php` (`Table` -> `TableFromCake`).

And then in `vendor/cakephp/cakephp/src/ORM/Table.php`:

```
<?php

namespace Cake\ORM;

class Table extends TableFromCake
{

    //book.cakephp.org/debugkit/5/en/index.html#tracing-query-execution
    use \DebugKit\Model\Table\SqlTraceTrait;

}
```

The downside is of course I did overwrite a vendor class. Maybe there is a way to have it done more elegantly than this. I first tried to [exclude it from composer](http://downing.tech/posts/overriding-vendor-classes) and then hijack it with [`class_alias()`](https://www.php.net/manual/en/function.class-alias.php), but couldn't come up with anything because 1. `Cake\ORM\Table` was still being autoloaded and 2. even if I did exclude it from autoload, then how do I import it back after defining the class alias without having it conflict with said class alias?.. Anyway, this gets the job done for now.

LordSimal:
The reason it was implemented this way was the fact, that this also works for `DELETE`, `INSERT` and `UPDATE` queries, not only for select queries.

If you only care for `SELECT` queries you can add a event listener to the `Model.beforeFind` event and add the logic from the `fileStamp` yourself. It maybe needs a few adjustments though since the stacktrace will not match what we expected in here.

To make this a global framework toggle and inject a comment on every query being executed I could think of adding a separate event for each query type (since [we already have separate classes for each query type](https://github.com/cakephp/cakephp/tree/5.x/src/ORM/Query)) and not only trigger a `Model.beforeFind` but also a e.g. `Model.beforeDelete`, `Model.beforeInsert` and `Model.beforeUpdate`.

LordSimal:
actually `Model.beforeDelete` already exists, but it doesn't contain the DeleteQuery instance, so that wouldn't work.
We'd have to add a separate event list to this kind of query hooking behavior.

mehov:
Hey Kevin, thank you for sharing your thoughts and for contributing this pull request. Just to be clear, I am perfectly happy with the hotfix solution that I got for now (rewriting the table class) and I'm not pushing/requesting for any changes on the framework level. I just shared my hotfix so that this shows up in search if someone else is looking to have something similar in development environment

LordSimal:
I think its a good idea to have in the backlog for future versions. 

Knowing exactly which SQL query is executed by which codeline can be very powerful in massive/ancient apps which may or may have not have been passed on from 20 different developers throwing patch fixes onto patch fixes for years.

And having just 1 toggle to switch instead of adding it per table can be very helpful.

PR #909, opened 2023-01-19
5.x: new stan setup

LordSimal:
Refs: https://github.com/cakephp/cakephp/pull/16953

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,5 +14,8 @@ trim_trailing_whitespace = true
 [*.neon]
 indent_style = tab
 
+[phars.xml]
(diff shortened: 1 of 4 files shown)


markstory:
Looks good to me, but I'm not well versed in using phive.

LordSimal:
All you should have to do is install phive on your local machine like explained [here](https://phar.io/) and then run `phive install` aka `composer stan-setup` to install the specified tools in `.phive/phars.xml` inside the `tools` directory.

On my machine I just had to make sure I had `gnupg` installed beause phive requires that. But other than that it should be pretty straight forward.

PR #930, opened 2023-04-25
4.x => 5.x merge

LordSimal:
had to add `require_once CAKE . 'functions.php';` to make `env()` available.
Or should this be handled differently?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ permissions:
 
 jobs:
   testsuite-linux:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
@@ -98,7 +98,7 @@ jobs:
 
   cs-stan:
     name: Coding Standard & Static Analysis
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


garas:
> had to add `require_once CAKE . 'functions.php';` to make `env()` available. Or should this be handled differently?

You should use namespaced `env()` instead

ADmad:
> had to add require_once CAKE . 'functions.php'; to make env() available.

As @garas said you need to use `Cake\Core\env()`. So add `use function Cake\Core\env` at the top of the required file.

LordSimal:
This would also mean I'd have to add the namespaced function for `h()` for all the template files which is not really what we want I guess (there are 62 usages of `h()` in 19 template files) - or am I wrong that we do want to go this way?

Especially since we made the [global functions opt-in](https://github.com/cakephp/cakephp/pull/17082) I'd still like to keep it the way it is right now since we only load those global functions for testing, not when the plugin is in use by an app.

ADmad:
Hmm.. Right, having to add `use` statement to the templates would be annoying.

LordSimal:
How about we automatically load the global functions in the rendering phase if they are not already present?
With that its not that hard of a BC but people can still overwrite/adjust them.

ADmad:
For the app the global functions are already loaded by default https://github.com/cakephp/app/blob/5.x/config/bootstrap.php#L52

LordSimal:
Yes, but what is the recommended way for plugin devs who use the global functions and tests which use them?

ADmad:
Plugins should ideally use the namespaced functions. But when its not convenient to do so like for templates here, we could include just the Core/functions_global.php file.

LordSimal:
adjusted it now to use the specific global_functions files which are actually used.

Issue #928, opened 2023-04-12
Introspect pagination results


dereuromark:
Currently, you have to pass the ResultSetDecorator into the view
->toArray() doesnt work, as the pagination needs those pagination data passed inside (at least in future versions for sure).
But that leads to this:

![vars](https://user-images.githubusercontent.com/39854/231403482-6b4cd8c3-513f-40c3-a4a1-32ef83ba2c7f.png)

It would be nice if there was an easy way to look into the actual results behind it instead of having to dd() or otherwise introspect the variable in the template (or controller) then manually.

LordSimal:
This is fixed with the next CakePHP 4.5 minor update. See linked PR's

PR #932, opened 2023-04-29
4.next: Introspect resultsets in variables panel

LordSimal:
Closes https://github.com/cakephp/debug_kit/issues/928

This will transform a variables panel like this
image

to this
image


dereuromark:
Should we also add the info that it at this time still is an object? Or will people know that we did this kind of Transformation for Display and debug purposes?

LordSimal:
Maybe I can find a different way to still output the items inside the ResultSetDecorator

LordSimal:
I found a better solution via adding a `__debugInfo()` method to the ResultSetDecorator. PR incoming for cake 4.next.

PR #934, opened 2023-04-29
4.x => 4.next merge and fix tests

LordSimal:


LordSimal:
The new `Doctrine\SqlFormatter\HtmlHighlighter` outputs PostgreSQL differently so we can't do a 1:1 compare like we did before.
See
```
Failed asserting that 
'/tests/TestCase/DebugSqlTest.php (line 73) SELECT panels.id AS &quot;panels__id&quot; FROM panels' 
contains 
"/tests/TestCase/DebugSqlTest.php (line 73) SELECT panels.id AS panels__id FROM panels".
```

PR #935, opened 2023-04-30
5.x: remove deprecations

LordSimal:
Haven't fully tried it with a test app but this removes the `ToolbarHelper::dump()` method which was marked as deprecated.

The tests working with `SimpleXmlElement` created some problems for me locally since the JS contained in the Debugger Header did not parse well as XML. Therfore I changed the test implementation to use `DOMDocument`.

Update: Nope, there are definitely some problems still present.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,56 +1,60 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.x-dev@">
+<files psalm-version="5.10.0@a5effd2d2dddd1a7ea7a0f6a051ce63ff979e356">
   <file src="src/DebugInclude.php">
-    <PossiblyNullArrayOffset occurrences="2">
-      <code>$this-&gt;_composerPaths</code>
-      <code>$this-&gt;_pluginPaths</code>
+    <PossiblyNullArrayOffset>
+      <code><![CDATA[$this->_composerPaths]]></code>
+      <code><![CDATA[$this->_pluginPaths]]></code>
     </PossiblyNullArrayOffset>
   </file>
   <file src="src/DebugSql.php">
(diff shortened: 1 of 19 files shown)


LordSimal:
I am probbaly misunderstanding things here but since we dropped `ToolbarHelper::dump()` one has to use `ToolbarHelper::dumpNodes()`

This one requires an array of nodes of type `\Cake\Error\Debug\NodeInterface` which can be created like
```
$nodeArray = [Debugger::exportVarAsNodes($request->getQueryParams())];
```

because `Debugger::exportVarAsNodes()` will always be 1 root Node containing all child nodes.

Therefore I had to do stuff like
```
'query' => [Debugger::exportVarAsNodes($request->getQueryParams())],
'data' => [Debugger::exportVarAsNodes($request->getData())],
'cookie' => [Debugger::exportVarAsNodes($request->getCookieParams())],
'get' => [Debugger::exportVarAsNodes($_GET)],
```

which feels wrong to me because it will created an output like this:
image

I seem to be missing something here so that we can also dump a single node. Should I just create that method in the ToolbarHelper?

markstory:
> Should I just create that method in the ToolbarHelper?

That seems like the correct approach. Accepting a list or single node in the existing method could also be a good approach.

LordSimal:
This looks better
image

Now there is only the issue of the duplicate copy button but I think this is something we need to fix in the core.

markstory:
Looks good!

LordSimal:
The duplicate copy buttons where happening because the debug_kit and cake core adds a separate `div` with the class `cake-debug`. But debug_kit doesn't need to do that anymore.

image

The copy button gets added to all `.cake-debug` divs via JS. Thats why it was duplicated.

PR #940, opened 2023-05-30
5.x: upgrade to PHPUnit 10

LordSimal:
Refs: https://github.com/cakephp/cakephp/pull/16992/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     "require-dev": {
         "cakephp/authorization": "3.x-dev",
         "cakephp/cakephp-codesniffer": "5.x-dev",
-        "phpunit/phpunit": "^9.5.19"
+        "phpunit/phpunit": "^10.1.0"
     },
     "autoload": {
         "psr-4": {
--- a/src/ToolbarService.php
+++ b/src/ToolbarService.php
@@ -102,7 +102,7 @@ class ToolbarService
      */
     public function isEnabled(): bool
     {
-        if (isset($GLOBALS['__PHPUNIT_BOOTSTRAP'])) {
+        if (defined('PHPUNIT_COMPOSER_INSTALL') && !$GLOBALS['FORCE_DEBUGKIT_TOOLBAR']) {
             return false;
         }
         $enabled = (bool)Configure::read('debug')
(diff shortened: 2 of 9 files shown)


composer.json
```diff
@@ -41,8 +41,8 @@
         "psr-4": {
             "Cake\\Test\\": "vendor/cakephp/cakephp/tests/",
             "DebugKit\\Test\\": "tests/",
-            "DebugKit\\TestApp\\": "tests/test_app/",
-            "DebugkitTestPlugin\\": "tests/test_app/Plugin/DebugkitTestPlugin/src/"
+            "DebugKit\\TestApp\\": "tests/TestApp/",
```
ADmad: I prefer having `test_app` over `TestApp`.

src/ToolbarService.php
```diff
@@ -102,7 +102,7 @@ public function registry(): PanelRegistry
      */
     public function isEnabled(): bool
     {
-        if (isset($GLOBALS['__PHPUNIT_BOOTSTRAP'])) {
+        if (defined('PHPUNIT_COMPOSER_INSTALL') && !$GLOBALS['FORCE_DEBUGKIT_TOOLBAR']) {
```
markstory, line 105: Of course they break my hacks :laughing:

LordSimal: Just re-introduced a new hack 😁

tests/TestCase/Middleware/DebugKitMiddlewareTest.php
```diff
@@ -125,7 +125,7 @@ public function testInvokeSaveData()
         $this->assertNotNull($result->panels[11]->summary);
         $this->assertSame('Sql Log', $result->panels[11]->title);
 
-        $timeStamp = filemtime(Plugin::path('DebugKit') . 'webroot' . DS . 'js' . DS . 'main.js');
+        $timeStamp = filectime(Plugin::path('DebugKit') . 'webroot' . DS . 'js' . DS . 'main.js');
```
garas, line 128: It should be `inject-iframe.js` here and keep using `filemtime`.

Line was previously incorrectly updated when .js file was renamed. And `ToolbarService::getToolbarUrl‎` is using `filemtime`.

LordSimal: I tried that but for some reason its not working... haven't looked too deep into it though.

PR #944, opened 2023-07-01
Use cspScriptNonce when it is available

### LINKED ISSUE #943: DebugKit not loading with CSP rule `strict-dynamic`
When I set up a fresh copy of CakePHP 4.4 and add the `CspMiddleware` in `Application.php` with the following settings ...

```
new CspMiddleware(['script-src' => ['strict-dynamic' => true]], ['scriptNonce' => true])
```

... the browser doesn't load DebugKit.

Error message:

> Refused to load the script 'http://localhost:8080/cakephp/debug_kit/js/inject-iframe.js?1674622148' because it violates the following Content Security Policy directive: "script-src 'nonce-24URd7b/3/ZF1wlidPdZCe53' 'strict-dynamic'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

Adding `'self' => true` to the `script-src` rule doesn't help as with `strict-dynamic` the host list gets ignored.

Adding a script via the HtmlHelper works flawlessly, because `HtmlHelper::script()` sets the nonce in the `<script>` tag correctly, whereas DebugKit's `ToolbarService::injectScripts()` doesn't take nonce into consideration.

I prefer to use the same CSP settings in development as in production to make sure everything works correctly, so I think DebugKit should also work with the CSP settings mentioned above.

markstory:
This allows debugkit to work in CSP controlled environments.

Fixes #943

### DIFF
--- a/tests/TestCase/Middleware/DebugKitMiddlewareTest.php
+++ b/tests/TestCase/Middleware/DebugKitMiddlewareTest.php
@@ -50,7 +50,6 @@ class DebugKitMiddlewareTest extends TestCase
         parent::setUp();
 
         $connection = ConnectionManager::get('test');
-        $this->skipIf($connection->getDriver() instanceof Sqlite, 'Schema insertion/removal breaks SQLite');
         $this->oldConfig = Configure::read('DebugKit');
         $this->restore = $GLOBALS['__PHPUNIT_BOOTSTRAP'];
         unset($GLOBALS['__PHPUNIT_BOOTSTRAP']);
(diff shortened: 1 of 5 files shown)


tests/TestCase/Middleware/DebugKitMiddlewareTest.php
```diff
@@ -50,7 +50,6 @@ public function setUp(): void
         parent::setUp();
 
         $connection = ConnectionManager::get('test');
-        $this->skipIf($connection->getDriver() instanceof Sqlite, 'Schema insertion/removal breaks SQLite');
```
markstory, line 53: This worked locally :crossed_fingers:

PR #945, opened 2023-07-02
5.x: update stan

LordSimal:


src/DebugMemory.php
```diff
@@ -62,7 +62,9 @@ public static function record(?string $message = null): bool
         $memoryUse = self::getCurrent();
         if (!$message) {
             $trace = debug_backtrace();
-            $message = Debugger::trimPath($trace[0]['file']) . ' line ' . $trace[0]['line'];
+            $file = $trace[0]['file'] ?? 'unknown file';
+            $line = $trace[0]['line'] ?? 'unknown line';
```
dereuromark: could also be n/a to not repeat line word.

dereuromark: Unknown file sounded good though

PR #947, opened 2023-07-28
Read cspScriptNonce from latest request

markstory:
Use Router::getRequest() to access the request from before controllers were dispatched and after middleware have been run. This should ensure that the CSP script nonce is present in the request.

Refs #943


othercorey:
Looks good.

PR #963, opened 2023-10-19
5.next: add Plugins Panel

LordSimal:
image

Refs: https://github.com/cakephp/cakephp/pull/17334


templates/element/plugins_panel.php
```diff
@@ -0,0 +1,50 @@
...
+    ?>
+    <section>
+        <table class="c-debug-table">
+            <thead>
+            <tr>
+                <th><?= __d('debug_kit', 'Plugin') ?></th>
+                <th><?= __d('debug_kit', 'Is Loaded') ?></th>
+                <th><?= __d('debug_kit', 'Only Debug') ?></th>
+                <th><?= __d('debug_kit', 'Only CLI') ?></th>
+                <th><?= __d('debug_kit', 'Optional') ?></th>
```
garas: No DebugKit localizations since 4.9.1

LordSimal: thanks, didn't notice that 😅

templates/element/plugins_panel.php
```diff
@@ -0,0 +1,50 @@
...
+                <th>Is Loaded</th>
+                <th>Only Debug</th>
+                <th>Only CLI</th>
+                <th>Optional</th>
+            </tr>
+            </thead>
+            <tbody>
+            <?php foreach ($plugins as $pluginName => $pluginConfig) : ?>
+                <tr>
+                    <td><?= $pluginName ?></td>
```
markstory:
```suggestion
                    <td><?= h($pluginName) ?></td>
```

Not likely to be necessary but we also don't need a security issue opened for a potential xss vector.

PR #971, opened 2023-10-30
5.x: fix elasticsearch logger error

### LINKED ISSUE #970: CakePHP 5 + Elasticsearch 4 causes an error "Call to undefined method Elastica\Client::getLogger()"
Hello,

When I install CakePHP 5 + Elasticsearch 4, I get an error:
`Call to undefined method Elastica\Client::getLogger()`

This error comes from the file:
`ROOT/vendor/cakephp/debug_kit/src/Panel/SqlLogPanel.php, line 60`

```php
$logger = $connection->getDriver()->getLogger();
```

Ubuntu 20.04.5 LTS (GNU/Linux 5.4.0-137-generic x86_64)
PHP 8.2.1

Install:
```
composer create-project --prefer-dist cakephp/app ./
composer require cakephp/elastic-search -W
```

To install Elasticsearch, you must pass the option --with-all-dependencies (-W), which results in a downgrading 
`react/promise (v3.0.0 => v2.10.0)`

It will also be necessary to add the configuration for elastic :

```php
'Datasources' => [
        'elastic' => [
            'className' => ElasticSearchConnection::class,
            'driver' => ElasticSearchConnection::class,
            'host' => '0.0.0.0',
            'port' => 9200,
            'transport' => 'https',
            'username' => 'elastic',
            'password' => 'verysecret',
        ],
    ],
```

Thanks

LordSimal:
Resolves #970 

It would be cool if @Christopher-HM could test this since I didn't want to do a whole elasticsearch setup 😅

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     "require-dev": {
         "cakephp/authorization": "^3.0",
         "cakephp/cakephp-codesniffer": "^5.0",
-        "phpunit/phpunit": "^10.1.0"
+        "phpunit/phpunit": "^10.1.0 <=10.5.3"
     },
     "autoload": {
         "psr-4": {
--- a/templates/element/sql_log_panel.php
+++ b/templates/element/sql_log_panel.php
@@ -76,7 +76,7 @@ $noOutput = true;
                         <tr>
                             <td>
                                 <?php
-                                if($query['isSqlQuery']):
+                                if ($query['isSqlQuery']):
                                     echo (new SqlFormatter(
                                         new HtmlHighlighter([
                                             HtmlHighlighter::HIGHLIGHT_QUOTE => 'style="color: #004d40;"',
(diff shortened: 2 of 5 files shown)


markstory (approved):

composer.json
```diff
@@ -30,7 +30,7 @@
     "require-dev": {
         "cakephp/authorization": "^3.0",
         "cakephp/cakephp-codesniffer": "^5.0",
-        "phpunit/phpunit": "^10.1.0"
+        "phpunit/phpunit": "^10.1.0 <=10.5.3"
```
LordSimal, line 33: Refs https://github.com/cakephp/cakephp/pull/17510

templates/element/sql_log_panel.php
```diff
@@ -75,8 +75,9 @@
                         <?php foreach ($queries as $query) : ?>
                         <tr>
                             <td>
-                                <?=
-                                    (new SqlFormatter(
+                                <?php
+                                if($query['isSqlQuery']):
```
ADmad:
```suggestion
                                if ($query['isSqlQuery']):
```

PR #975, opened 2023-11-27
Run composer normalize over composer.json

dereuromark:
Auto created via toolbox app.

- Runs https://github.com/ergebnis/composer-normalize over composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,10 +1,14 @@
 {
     "name": "cakephp/debug_kit",
     "description": "CakePHP Debug Kit",
-    "type": "cakephp-plugin",
-    "keywords": ["cakephp", "debug", "kit", "dev"],
-    "homepage": "https://github.com/cakephp/debug_kit",
     "license": "MIT",
+    "type": "cakephp-plugin",
+    "keywords": [
+        "cakephp",
+        "debug",
+        "kit",
+        "dev"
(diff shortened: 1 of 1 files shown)


dereuromark:
If this is something we want to do to normalize it across all repos, I can run the tool over all relevant repos.
Just let me know.

There might be more merge conflicts at the start, but then should minimize afterwards.

ravage84 (commented):
While I do like uniformity most often, I'm not sure I like it in this "opinionated" way. 🤷

dereuromark:
Is it? Afaik it aligns itself according to the underlying [JSON schema](https://getcomposer.org/schema.json)
So it just follows what composer dictates as Schema it seems.

othercorey:
Looks ok to me. I remember doing this manually once, but I guess I never committed.

LordSimal:
The only thing that's weird to me is the position of the `suggest` block. Everything is fine by me 👍🏻 
But I guess we can all get used to it. Would love to see it across all our repos AND a github CI check

dereuromark:
Well, we can reach out to composer and ask them to fix up the order as per our suggestions - pun intended :)
Before we further modify more repos.

As for the CI check: Might be a bit overkill, so often we will not change the structure as a whole.

PR #976, opened 2023-11-28
Add external panel info

dereuromark:
See https://github.com/dereuromark/cakephp-setup/blob/master/docs/Panel/L10nPanel.md

L10n

Do we want to add here other external panels for easier showcasing and people being able to pull them in?


markstory:
> Do we want to add here other external panels for easier showcasing and people being able to pull them in?

I think including links to plugins that have debug kit integrations makes sense.

PR #979, opened 2023-12-15
Fixed ReflectionProperty::setValue() deprecation warning when working on PHP 8.3

Bird87ZA:
On PHP 8.3 you get this deprecation warning:

image

I've changed the deprecated method out with the correct method. All tests (non -skipped) pass, and I can't see any issues.


othercorey:
Thanks! We should backport this to 4.x.

Bird87ZA:
If needed I'll open another PR for it shortly

othercorey:
If you can, that would be helpful.

PR #981, opened 2023-12-21
Fix for deprecation warning in SimpleGraphHelper::bar()

jharder:
When using DebugKit, on the Timer tab the bar graphs all display an error message:

`Deprecated (8192): Implicit conversion from float 126.51705741882324 to int loses precision [in /var/cake/vendor/cakephp/debug_kit/src/View/Helper/SimpleGraphHelper.php, line 55]`

The bar() function currently uses these parameters like this:
`$value`:
`$graphValue = $value / $max * $width;`

`$offset`:
`$graphOffset = $offset / $max * $width;`

Both new `$graph*` values are then provided to round(), which handles conversion to int.  The raw values are then only used in displaying the message. From what I can tell, this function can handle float as well as int values, and this fix updates the type hint to allow the actual consumers of the function to use it as they have historically. 

Added a new test to cover the explicit case. All test cases pass.

### DIFF
--- a/src/View/Helper/SimpleGraphHelper.php
+++ b/src/View/Helper/SimpleGraphHelper.php
@@ -47,12 +47,12 @@ class SimpleGraphHelper extends Helper
     /**
      * bar method
      *
-     * @param float $value Value to be graphed
-     * @param int $offset how much indentation
+     * @param float|int $value Value to be graphed
+     * @param float|int $offset how much indentation
      * @param array $options Graph options
      * @return string Html graph
      */
-    public function bar(float $value, int $offset, array $options = []): string
+    public function bar(float|int $value, float|int $offset, array $options = []): string
     {
         $settings = array_merge($this->_defaultSettings, $options);
         $max = $settings['max'];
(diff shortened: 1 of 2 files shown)


src/View/Helper/SimpleGraphHelper.php
```diff
@@ -47,12 +47,12 @@ class SimpleGraphHelper extends Helper
...
      *
-     * @param float $value Value to be graphed
-     * @param int $offset how much indentation
+     * @param float|int $value Value to be graphed
+     * @param float|int $offset how much indentation
      * @param array $options Graph options
      * @return string Html graph
      */
-    public function bar(float $value, int $offset, array $options = []): string
+    public function bar(float|int $value, float|int $offset, array $options = []): string
```
dereuromark, line 55: Why do we need this? Couldnt we just cast to int beforehand?

jharder: Then you would lose precision in the text output of the alt text "Starting 0ms into the request, taking 263.582ms".  If we cast to int before calling this function, then probably a bunch of the times would round down to 0ms, which is incorrect. 

As far as I can tell there is only one caller to this method, and that's in `templates/element/timer_panel.php`.  You could cast to int there, but looking at a my own time graphs, there are quite a few elements that would lose information by reducing the input precision.

PR #993, opened 2024-03-02
Add route path for easier use.

### LINKED ISSUE #992: Feature: Route path on request tab
Currently, to know where this URL maps to, one has to open request tab and click
on 

    Attributes -> Params

Wouldnt it be nice if this info would be available at the top, e.g. in form of a named route string?
https://book.cakephp.org/5/en/development/routing.html#using-path-routing

    [Plugin].[Prefix]/[Controller]::[action]

This way, one can also more easily copy and paste this for use in link()/url() calls or other ways (e.g. ACL/Auth, ...

dereuromark:
Resolves https://github.com/cakephp/debug_kit/issues/992

debug_kit

### DIFF
--- a/src/Panel/RequestPanel.php
+++ b/src/Panel/RequestPanel.php
@@ -49,6 +49,7 @@ class RequestPanel extends DebugPanel
         }
 
         $this->_data = [
+            'params' => $request->getAttribute('params'),
             'attributes' => $attributes,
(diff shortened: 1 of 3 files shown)


dereuromark:
Looks like https://github.com/cakephp/debug_kit/pull/994 needs to get merged first

PR #995, opened 2024-03-20
Add DOMContentLoaded event alternative for Turbo

sergioalcaraz:
In this pull request, the turbo:load event is added together with the already existing DOMContentLoaded event to avoid having to reload the page on each navigation.

### DIFF
--- a/webroot/js/inject-iframe.js
+++ b/webroot/js/inject-iframe.js
@@ -98,9 +98,12 @@ if (elem) {
   if (doc.addEventListener) {
     // This ensures that all event listeners get applied only once.
     if (!win.debugKitListenersApplied) {
-      doc.addEventListener('DOMContentLoaded', onReady, false);
-      doc.addEventListener('DOMContentLoaded', proxyAjaxOpen, false);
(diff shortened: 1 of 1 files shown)


webroot/js/inject-iframe.js
```diff
@@ -101,6 +101,12 @@ if (elem) {
       doc.addEventListener('DOMContentLoaded', onReady, false);
       doc.addEventListener('DOMContentLoaded', proxyAjaxOpen, false);
       doc.addEventListener('DOMContentLoaded', proxyAjaxSend, false);
+
+      // Add support for turbo DOMContentLoaded alternative
+      // see https://turbo.hotwired.dev/reference/events#turbo%3Aload
+      doc.addEventListener('turbo:load', onReady, false);
```
markstory: This will close the toolbar and attach duplicate message handlers. I don't think that's what we want.

sergioalcaraz: Added a condition to set which of the two events to be used

Turbo ? 'turbo:load' : 'DOMContentLoaded';

markstory:
Thank you 🎉

PR #1004, opened 2024-05-27
Add empty state for deprecations

markstory:
Make it more obvious that the panel isn't broken.

### DIFF
--- a/templates/element/deprecations_panel.php
+++ b/templates/element/deprecations_panel.php
@@ -23,7 +23,7 @@
 
 use function Cake\Core\h;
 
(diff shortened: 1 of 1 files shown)


templates/element/deprecations_panel.php
```diff
@@ -23,6 +23,8 @@
 
 use function Cake\Core\h;
 
+$hasAny = count($app) + count($plugins) + count($cake) . count($vendor);
```
garas:
```suggestion
$hasAny = count($app) + count($plugins) + count($cake) + count($vendor);
```

markstory: Oops. Thanks for the edit.

PR #1014, opened 2024-08-08
update stan

LordSimal:


dereuromark (approved):

src/View/Helper/CredentialsHelper.php
```diff
@@ -50,7 +50,7 @@ public function filter(mixed $in): mixed
             return $in;
         }
         preg_match_all($regexp, $in, $tokens);
-        if (empty($tokens[0])) {
+        if (count($tokens[0]) === 0) {
```
dereuromark: is that a PHPStan requirement? sounds a bit silly to require the count

LordSimal: its a psalm requirement.
```
ERROR: RiskyTruthyFalsyComparison
at /Users/kevin/Documents/CakePHP/Org/debug_kit/src/View/Helper/CredentialsHelper.php:53:13
Operand of type list<string>|null contains type list<string>, which can be falsy and truthy. This can cause possibly unexpected behavior. Use strict comparison instead. (see https://psalm.dev/356)
        if (empty($tokens[0])) {
```

maybe there is a better comparison but this was just a quickfix to get it green again

ADmad:
```suggestion
        if ($tokens[0] === []) {
```
If the index is always going to be set.

ADmad: For the above regex it will always be set.

PR #1015, opened 2024-08-08
allow PHPUnit 11 and adjust tests for PHPUnit 11

LordSimal:


tests/TestCase/Panel/VariablesPanelTest.php
```diff
@@ -69,16 +69,15 @@ public function testShutdown()
         $update = $requests->updateQuery();
         $debugInfoException = $requests->query()->contain('NonExistentAssociation');
 
-        $unserializable = new stdClass();
-        $unserializable->pdo = new PDO('sqlite::memory:');
+        $unserializableDebugInfo = new class extends stdClass {
```
markstory, line 72: I like this approach much better than mocks. We should do more of this :clap:

Issue #1021, opened 2024-08-19
When debugkit SQLite size is after some size it's never cleared again


Harfusha:
When debugkit SQLite file size is bigger than 500 MB (On our system), rows are deleted, but vacuum is never called successfully.

When we tried to call it manualy this error occured: `sqlite> VACUUM; Error: out of memory`

Maybe instead of just `VACUUM;` should be called `VACUUM INTO {FILE};` to reduce memory requirements. 

And another problem: When we were debugging this issue we added counter of requests between garbage collection and it normally reached ~1000 requests and in this moment the database was over 2GB in size -> so VACUUM throwed out of memory error

LordSimal:
If you don't care about the data inside debug_kit (so saved request data) then you can just remove that `tmp/debug_kit.sqlite` and you are fine. It gets automatically recreated.

I have not yet run into this problem. My current `tmp/debug_kit.sqlite` is 223MB big and it vacuums just fine.

Garbage collection/vacuum happens automatically after around 100 requests as you [can see here](https://github.com/cakephp/debug_kit/blob/5.x/src/Model/Table/RequestsTable.php#L87-L90)

Also you can configure how many requests are being preserved while doing GC by setting `DebugKit.requestCount` to e.g. `10` instead of the [default 20](https://github.com/cakephp/debug_kit/blob/5.x/src/Model/Table/RequestsTable.php#L112C42-L112C63)

If you don't want SQLite to be used for debug kit then you can easily just [configure another DBMS](https://book.cakephp.org/debugkit/5/en/index.html#database-configuration) to be used for saving request data for debug kit

Harfusha:
But problem is that is is growing until no more space on server is left

> Garbage collection/vacuum happens automatically after around 100 requests as you [can see here](https://github.com/cakephp/debug_kit/blob/5.x/src/Model/Table/RequestsTable.php#L87-L90)
I know but when the issue i described happened it was 1156 requests from last GC

LordSimal:
The Debug Kit is only meant to be used when developing the website (aka in your local setup), not on your live website/server.

And besides: Not enough diskspace is not the same as your reported `out of memory` problem.

Harfusha:
I know that 1% chance of GC should call it ~1 in 100 requests, but when it doesnt the file can get to size where it causes issues. Shouldn't there be some hardlimit when to call GC forcefully (without the 1% hit?).

[...]

PR #1022, opened 2024-08-22
Add deflate and inflate of panel content, changed gc chance from 1% t…

Harfusha:
…o 10%, added 10% chance to vacuum when sqlite is used, added pragma commands for optimalization when using sqlite

https://github.com/cakephp/debug_kit/issues/1021

### DIFF
--- a/src/Model/Entity/Panel.php
+++ b/src/Model/Entity/Panel.php
@@ -72,7 +72,7 @@ class Panel extends Entity
     {
         if (is_string($content) && function_exists('gzdeflate')) {
             $contentDeflated = gzdeflate($content, 9);
-            if ($contentDeflated) {
+            if ($contentDeflated !== false) {
                 $content = $contentDeflated;
             }
         }
--- a/src/Model/Entity/Panel.php
+++ b/src/Model/Entity/Panel.php
@@ -47,8 +47,10 @@ class Panel extends Entity
     protected function _getContent(mixed $content): string
     {
         if (is_string($content)) {
+            // phpcs:disable
             $contentInflated = @gzinflate($content);
-            if ($contentInflated) {
+            // phpcs:enable
+            if ($contentInflated !== false) {
                 return $contentInflated;
             }
 
@@ -72,7 +74,7 @@ class Panel extends Entity
     {
         if (is_string($content)) {
             $contentDeflated = gzdeflate($content, 9);
-            if ($contentDeflated) {
+            if ($contentDeflated !== false) {
                 $content = $contentDeflated;
             }
         }
--- a/src/Model/Entity/Panel.php
+++ b/src/Model/Entity/Panel.php
@@ -36,7 +36,7 @@ class Panel extends Entity
     protected array $_hidden = ['content'];
 
     /**
-     * Read the stream contents.
+     * Read the stream contents or inflate deflated data.
      *
      * Over certain sizes PDO will return file handles.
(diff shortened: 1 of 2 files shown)


src/Model/Entity/Panel.php
```diff
@@ -46,10 +46,37 @@ class Panel extends Entity
      */
     protected function _getContent(mixed $content): string
     {
+        if (is_string($content)) {
+            $contentInflated = @gzinflate($content);
+            if ($contentInflated) {
```
markstory:
```suggestion
            if ($contentInflated !== false) {
```

comparing to false will avoid empty string from being ignored.

src/Model/Entity/Panel.php
```diff
@@ -46,10 +46,37 @@ class Panel extends Entity
      */
     protected function _getContent(mixed $content): string
     {
+        if (is_string($content)) {
+            $contentInflated = @gzinflate($content);
```
LordSimal, line 55: According to https://www.php.net/manual/en/zlib.installation.php zlib is not installed by default via PHP.

To not force all users to install the php zlib extension I'd simply wrap that in a 
```
if(function_exists('gzinflate'))
``` 
to make sure this is actually present.

src/Model/Entity/Panel.php
```diff
@@ -46,10 +46,37 @@ class Panel extends Entity
...
+     * Deflate the string data before saving it into database
+     *
+     * @param mixed $content Content
+     * @return mixed
+     */
+    protected function _setContent(mixed $content): mixed
+    {
+        if (is_string($content)) {
+            $contentDeflated = gzdeflate($content, 9);
+            if ($contentDeflated) {
```
markstory:
```suggestion
            if ($contentDeflated !== false) {
```

Same here.

src/Model/Table/RequestsTable.php
```diff
@@ -131,7 +141,26 @@ public function gc(): void
...
             if ($conn->getDriver() instanceof Sqlite) {
-                $conn->execute('VACUUM;');
+                $conn->execute('
+                    PRAGMA auto_vacuum = FULL;
+                    PRAGMA journal_mode = OFF;
+                    PRAGMA synchronous = OFF;
+                    PRAGMA foreign_keys = OFF;
+                    PRAGMA temp_store = MEMORY;
+                    PRAGMA automatic_index = OFF;
+                ');
```
markstory, line 152: Should this only happen if the GC is going to occur?

src/Model/Table/RequestsTable.php
```diff
@@ -131,7 +141,26 @@ public function gc(): void
...
+
+                if (!$this->shouldGcVacuum()) {
```
markstory, line 154: If we've already passed the rand to do a GC why do we need another rand to do vacuum?

src/Model/Table/RequestsTable.php
Harfusha: I think it is not needed to run everytime, so it is run when gc is ran

src/Model/Table/RequestsTable.php
Harfusha: So rows are cleared more frequently, but the slow vacuum should be ran only 1% of requests (I made the check to own functions for static analysis tools and if we want to tweak the chance)

Harfusha:
Any other issues with this?

PR #1027, opened 2024-09-17
fix styling for environment panel

LordSimal:
image
image

thanks @markstory for pointing into the right direction 😉 
this of course is just a debug_kit specific CSS problem, not a debugger general problem


markstory:
Thanks for doing this. Simple improvements make all the difference over time.

PR #1029, opened 2024-11-17
update stan

LordSimal:


src/Panel/RoutesPanel.php
```diff
@@ -40,16 +40,14 @@ public function summary(): string
...
      * Data collection callback.
      *
-     * @param \Cake\Event\EventInterface $event The shutdown event.
+     * @param \Cake\Event\EventInterface<\Cake\Controller\Controller> $event The shutdown event.
      * @return void
      */
     public function shutdown(EventInterface $event): void
     {
-        /** @var \Cake\Controller\Controller|null $controller */
         $controller = $event->getSubject();
```
LordSimal, line 48: This line either must return the controller object or throws an exception, if no subject is present. Therefore the whole nullable checks are not needed anymore.

src/Panel/SqlLogPanel.php
```diff
@@ -52,10 +51,7 @@ public function initialize(): void
 
         foreach ($configs as $name) {
             $connection = ConnectionManager::get($name);
-            if (
-                $connection->configName() === 'debug_kit'
-                || !$connection instanceof ConnectionInterface
-            ) {
+            if ($connection->configName() === 'debug_kit') {
```
ADmad, line 54: Hmm.. 4.x allowed the `ConnectionManager` to hold connection instances which did not implement `ConnectionInterface`, for e.g. https://github.com/UseMuffin/Webservice/blob/6376dc722529d888c4750793f0b286a8acd9fd71/src/Datasource/Connection.php#L19.

Guess we killed that in 5.0 by adding `ConnnectionInterface` as return type for `ConnectionManager::get()`.

PR #1033, opened 2024-12-29
add connection role to SQL Log Panel queries

### LINKED ISSUE #1003: Read connections in SQL Log panel?
I recently stumbled across [this](https://book.cakephp.org/5/en/orm/database-basics.html#read-and-write-connections) great feature that allows read-only connections, but I do not see any queries using the "read" connection in the SQL Log panel within DebugKit.

Does DebugKit support these "read" connections? Perhaps I'm missing some configuration setting?

The panel code seems to look for unique connections, in this cause it finds "default", but the "read" and "write" config keys specified seem to both be a part of "default", so they're not seen as two different connections. The end result is that the SQL Log panel only shows "write" connections, no "read" connections are visible.

```
'default' => [
	'className' => Connection::class,
	'driver' => Postgres::class,
	'encoding' => 'utf8',
	'host' => 'localhost',
	'username' => 'test',
	'password' => 'test',
	'database' => 'test_db',
	'read' => [
		'database' => 'read_test_db',
	],
	'write' => [
		'database' => 'test_db',
	],
	'log' => false,
],
```

Here is the original PR for this feature: https://github.com/cakephp/cakephp/pull/16785

LordSimal:
Fixes https://github.com/cakephp/debug_kit/issues/1003

image

### DIFF
--- a/src/Database/Log/DebugLog.php
+++ b/src/Database/Log/DebugLog.php
@@ -134,6 +134,7 @@ class DebugLog extends AbstractLogger
      */
     public function log($level, string|Stringable $message, array $context = []): void
     {
+        /** @var \Cake\Database\Log\LoggedQuery|object|null $query */
         $query = $context['query'] ?? null;
 
         if ($this->_logger) {
@@ -153,6 +154,9 @@ class DebugLog extends AbstractLogger
                 ], JSON_PRETTY_PRINT),
                 'took' => $took,
                 'rows' => $context['response']['hits']['total']['value'] ?? $context['response']['hits']['total'] ?? 0,
+                'inTransaction' => $this->inTransaction,
+                'isCommitOrRollback' => false,
+                'role' => '',
             ];
 
             return;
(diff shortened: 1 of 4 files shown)


src/Database/Log/DebugLog.php
```diff
@@ -153,6 +154,9 @@ public function log($level, string|Stringable $message, array $context = []): vo
                 ], JSON_PRETTY_PRINT),
                 'took' => $took,
                 'rows' => $context['response']['hits']['total']['value'] ?? $context['response']['hits']['total'] ?? 0,
+                'inTransaction' => $this->inTransaction,
+                'isCommitOrRollback' => false,
+                'role' => '',
```
LordSimal, line 159: This was added to make it consistent for Elastic Search query logging as well

LordSimal:
@ryanolton would this be what you wanted?

ryanolton:
@LordSimal yes, that looks fantastic, thanks!

PR #1038, opened 2025-06-18
POC Dump panel.

dereuromark:
Normal debug() breaks templates and JS, so with debug kit enabled this should go into dump panel IMO

Anywhere:
```php
debugKitDump('fpopbar');

...

debugKitDump('rrrr');
```

Proof of concept.

Before
before

After
a

and
after


dereuromark:
Downside of own method:
If debugkit is not loaded, this will fail hard.

Using debug() should be preferred, but my tests revealed that core one is loaded first, so not sure how to fix that up.
When locally debugging, I think it is fine usually to require this method to be removed again after usage.

LordSimal:
I have no strong opinions about this, since I wouldn't really use this feature.
Besides the fact that it breaks the usual PHP debugging convention, where the debug output is printed where it is executed.

But if people like to have more options to do that, I am fine with that as well.

dereuromark:
File and line is still visible, same as logging it always is

templates/element/dump_panel.php
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @var \DebugKit\View\AjaxView $this
+ * @var array $statements
+ */
+
+?>
+<div class="c-dump-panel">
+    <?php
+    foreach ($statements as $statement) {
+        echo '<pre class="cake-debug-string">' . h($statement['var']) . '</pre>';
```
markstory, line 26: Don't we need to handle more complex values as well? If we're only going to support simple values, one could use logging instead.

markstory (commented):
Since you want to defer output, why would someone use this over `Log`?

PR #1040, opened 2025-07-16
Add static API to add connections to the SqlPanel after initialization.

ADmad:


dereuromark:
In the panel output, in the last row "Role" we could add below the connection it came from.
Or we could split it at the top using tabs or alike to filter by a specific connection?
I wonder how else we could make it visible/clear what queries ran on which one.

//EDIT
OK, they are below each other, but the name is on the top of each block.
SO guess thats fair enough for now

Filtering could be added later.

LordSimal (approved):

PR #1045, opened 2025-09-13
Make DebugKit toolbar robust with JS frameworks like HTMX, Hotwire, ...

josbeir:
This PR makes the  toolbar work when used with  libraries (such as HTMX, Hotwire, etc.) that swap or
replace the `<body>` element or perform partial page updates. 

### Improvements

*  The toolbar is now kept outside of the documet.body, this keeps it from being removed, even if the body element is swapped.

**The main error i experienced was** 

```
Uncaught TypeError: can't access property "postMessage", iframe.contentWindow is null
```

This error occured after transitioning to another page (e.g., with htmx:boost or similar techniques) because the toolbar iframe was lost or not properly re-injected.

### DIFF
--- a/webroot/js/inject-iframe.js
+++ b/webroot/js/inject-iframe.js
@@ -10,27 +10,18 @@ if (elem) {
   let bodyOverflow;
 
   const onMessage = (event) => {
-    if (event.data === 'collapse') {
-      iframe.height = 40;
-      iframe.width = 40;
-      doc.body.style.overflow = bodyOverflow;
-      return;
-    }
-    if (event.data === 'toolbar') {
-      iframe.height = 40;
-      iframe.width = '100%';
-      doc.body.style.overflow = bodyOverflow;
-      return;
-    }
-    if (event.data === 'expand') {
-      iframe.width = '100%';
-      iframe.height = '100%';
-      doc.body.style.overflow = 'hidden';
-    }
-    if (event.data === 'error') {
-      iframe.width = '40%';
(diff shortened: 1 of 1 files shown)


josbeir:
This is now ready for testing / comments :-)

josbeir:
I changed some things, realizing I overcomplicated it by adding the iframe to the shadow root, which I then moved outside document.body. The iframe alone outside of document.body is enough here.

markstory:
> The iframe alone outside of document.body is enough here.

Is that a valid location for iframes? I tried reading through the HTML spec and iframes are part of 'embedded content' and I'm not sure that iframes can be under the `html` element.

Another solution could be to have debugkit render for these requests. We could expose it as a setting. I don't think we can start appending debugkit to all HTML responses, without an opt-in.

josbeir:
> > The iframe alone outside of document.body is enough here.
> 
> Is that a valid location for iframes? I tried reading through the HTML spec and iframes are part of 'embedded content' and I'm not sure that iframes can be under the `html` element.
> 
Browsers are pretty forgiving about this but after some extra research it looks like this is indeed not really valid if you follow the html spec stricly.  You can therefore currently ignore this MR. 

> Another solution could be to have debugkit render for these requests. We could expose it as a setting. I don't think we can start appending debugkit to all HTML responses, without an opt-in.

Yeah i think a more dynamic approach would be better anyway. I will close this MR and do some thinking for a cleaner solution.

PR #1053, opened 2025-11-02
update stan & cleanup tests

LordSimal:


composer.json
```diff
@@ -30,13 +30,13 @@
     "require": {
         "php": ">=8.1",
         "cakephp/cakephp": "^5.1",
-        "composer/composer": "^2.0",
+        "composer/composer": "^2.7.7",
```
LordSimal, line 33: There were security updates for composer
https://getcomposer.org/changelog/2.7.7

Issue #821, opened 2021-03-16
Handle out of memory on serialization better


othercorey:
Whenever there is a type that can't be serialized, `ToolbarService` throws a confusing out of memory fatal error.

Most users don't know what `ToolbarService` is trying to do which makes it hard to debug.

![image](https://user-images.githubusercontent.com/24221186/111258536-32249880-85eb-11eb-97fb-be15a905e7cc.png)

tbelknapsgs:
It is extremely confusing! In fact, the DebugKit itself needs some debugging work: any time it gets data it doesn't understand it seems to dump with an error. For example: I incorrectly formatted my Entity class's $_virtual array (I thought it was supposed to look like the $_accessible array) and I got error messages from DebugKit about an Entity formatting issue but it took forever to figure out that it was DebugKit, not CakePHP proper, choking on my error.

LordSimal:
@tbelknapsgs sorry for taking so long to respond, but I am currently trying to reproduce this error.

I tried to add 
```
protected array $_virtual = ['first_name' => false];
```
to my entity since this is exactly what you described, but it doesn't produce the error above.

Could you maybe explain what needs to be done to break debug kit?

LordSimal:
partially fixed by the attached PR, but running into memory limits while serialization is being executed will still cause weird errors.

LordSimal:
As there were no further comments from other users as well as no reproducable way to debug this issue I will close this.

PR #1054, opened 2025-11-03
Try to reproduce serialize issues

dereuromark:
and avoid them with a better handling if possible.

Refs https://github.com/cakephp/debug_kit/issues/821


LordSimal:
Looks good to me, just needs to fix CS and that test where the panel doesn't actually exist (thanks AI)

dereuromark:
Feel free to take over or borrow the code that seems useful and close. I am currently on the migrations topic.

PR #1055, opened 2025-11-14
Allow debug mode by default.

dereuromark:
Whats the whole point of this flag if that action is only to be used in debug mode

 > Cake\Http\Exception\NotFoundException: Not available without debug mode o

And debug mode by definition should only be a local dev thing (thus access should be granted).

Having to set this flag for all apps is quite annoying, and I dont really see any security relevant point in it.

For now you can still disable using

    Configure::write('DebugKit.ignoreAuthorization', false);

but I dont quite see the point here for DebugKit and debug mode on (the only way it can actually work and display things anyway).

### DIFF
--- a/src/Controller/DebugKitController.php
+++ b/src/Controller/DebugKitController.php
@@ -43,7 +43,7 @@ class DebugKitController extends Controller
         // ignore it, only if `DebugKit.ignoreAuthorization` is set to true
         $authorizationService = $this->getRequest()->getAttribute('authorization');
         if ($authorizationService instanceof AuthorizationService) {
-            if (Configure::read('DebugKit.ignoreAuthorization') || Configure::read('debug')) {
+            if (Configure::read('DebugKit.ignoreAuthorization') !== false) {
                 $authorizationService->skipAuthorization();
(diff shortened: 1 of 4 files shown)


src/Controller/DebugKitController.php
```diff
@@ -43,7 +43,7 @@ public function beforeFilter(EventInterface $event): void
         // ignore it, only if `DebugKit.ignoreAuthorization` is set to true
         $authorizationService = $this->getRequest()->getAttribute('authorization');
         if ($authorizationService instanceof AuthorizationService) {
-            if (Configure::read('DebugKit.ignoreAuthorization')) {
+            if (Configure::read('DebugKit.ignoreAuthorization') !== false) {
```
markstory, line 46: We had defaulted to off to take care of the inevitable situation of someone deploying with debug on. The hope was that folks wouldn't also deploy this config option.

dereuromark: Thats highly unlikely, as this is usually a committed config setting then (not a app_local one) :)
So not really protecting anyone here IMO

PR #1062, opened 2026-03-14
rector

LordSimal:
This would be changed by rector if we added it to debug_kit. Would love to see this across all our plugins to get more consistency.

Woud be cool if we could somehow unify the rector config across repositories though.

The shared CI workflow also needs to be adjusted to execute rector (if its present)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,6 +66,9 @@
         "stan": "@phpstan",
         "stan-baseline": "tools/phpstan --generate-baseline",
         "stan-setup": "phive install",
+        "rector-setup": "cp composer.json composer.backup && composer require --dev rector/rector:\"~2.3.1\" && mv composer.backup composer.json",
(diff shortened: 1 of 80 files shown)


LordSimal:
Refs: https://github.com/cakephp/.github/pull/26

markstory (commented):
The changes rector applied seem good to me. Would we be able to put `rector.php` into   a package that we can share across plugins?

LordSimal:
A shared (or at least baseline) rector config would be great, but we'd need to add it to something all projects depend on.

The rector config is just a object with chainable config calls, so it would be just 1 additional require before the rector.php in each package using it.

We could of course also just make an additional package/repo and add it as a dev dependency, but I don't know if this is the best way to go.

markstory (approved):

PR #1074, opened 2026-05-13
JS DOM-safe rendering: stop string-concat HTML into the toolbar iframe

dereuromark:
## Summary

Third followup PR from the 5.x audit -- the JS hygiene pass.

The toolbar iframe is loaded **same-origin** with the host app, so any HTML injection inside it is effectively XSS against the app being developed (cookies, CSRF tokens, in-flight session). Several panels were composing HTML by string concatenation from data that crosses the toolbar's trust boundary (cross-origin response `Content-Type` headers, attacker-influenced URLs, raw composer error bodies). Each sink is rewritten to construct DOM via `document.createElement` / jQuery `.text()` / `.append()` instead.

### What changed

* **HistoryPanel.js** -- previously the per-XHR row was built by string-replacing six placeholders (`{id}`, `{time}`, `{method}`, `{status}`, `{type}`, `{url}`) into raw HTML and inserting the result with `.after(htmlString)`. `params.type` is the response `Content-Type` of any cross-origin URL the dev's app fetches and is therefore attacker-controlled; `params.url` and `params.method` are similarly untrusted. New flow: substitute only the (sanitized) request UUID `{id}` into the href and `data-request` attributes, then populate the five visible spans via jQuery `.text()`. The `[data-request=...]` highlight selector at the bottom of `init()` is rewritten as a `.filter(fn)` predicate so a runtime id can't slip into selector syntax.

* **CachePanel.js / RequestPanel.js** -- `$(`<p>${text}</p>`)` -> `$('<p>').text(text)`. Same UX, no HTML parsing.

* **PackagesPanel.js** -- the success / error output is assembled as jQuery-built elements with `.text()`, and `showMessage()` writes them via `.empty().append(...)` rather than `.html(string)`. Fixes a pre-existing bug along the way: the `error` callback was passing jQuery's `textStatus` string to `buildErrorMessage` (which expected the jqXHR object), so every non-2xx response was previously a silent `JSON.parse` throw with no visible message. The new error builder takes the jqXHR object directly and falls back to `responseText` / `statusText` / a literal `"Request failed"` -- credit to codex for catching the regression.

* **postMessage handlers** in `inject-iframe.js` (parent side) and `Toolbar.onMessage` (iframe side) now reject any message whose origin is not same-origin and whose source is not the expected counterpart window (`iframe.contentWindow` on the parent, `window.parent` on the iframe). Today only resize and XHR-counter messages flow, so the practical impact of the missing check is low, but the one-line hardening keeps the protocol safe as it grows.

### Verification

ESLint output for the touched files is clean; the remaining warnings/errors (`Turbo` undefined, unnamed function expressions in Toolbar and inject-iframe, Keyboard.js return) are pre-existing on 5.x and untouched by this PR. PHP test suite, PHPStan, and PHPCS all green.

### DIFF
--- a/webroot/js/inject-iframe.js
+++ b/webroot/js/inject-iframe.js
@@ -10,6 +10,14 @@ if (elem) {
   let bodyOverflow;
 
   const onMessage = (event) => {
+    // The toolbar iframe is same-origin with the host app; reject any
+    // postMessage that does not originate from the iframe we created.
+    if (event.origin !== win.location.origin) {
+      return;
+    }
+    if (!iframe || event.source !== iframe.contentWindow) {
+      return;
+    }
     if (event.data === 'collapse') {
       iframe.height = 40;
       iframe.width = 40;
--- a/webroot/js/modules/Panels/CachePanel.js
+++ b/webroot/js/modules/Panels/CachePanel.js
@@ -1,6 +1,6 @@
 export default (($) => {
   const addMessage = (text) => {
-    $(`<p>${text}</p>`)
+    $('<p>').text(text)
       .appendTo('.c-cache-panel__messages')
       .fadeOut(2000);
   };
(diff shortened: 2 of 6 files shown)


LordSimal (approved):
LGTM

PR #1076, opened 2026-05-25
Fix TypeError: Cannot read properties of null

sergioalcaraz:
Turbo expects that when there are errors in validations, for example, it should respond with status 422, which causes the iframe to disappear from the DOM, so the null error occurs in subsequent requests.


markstory:
> Turbo expects that when there are errors in validations, for example, it should respond with status 422, which causes the iframe to disappear from the DOM

If the iframe is gone won't debugkit be broken?

LordSimal:
I think his point is the fact, that there shouldn't be additional error messages when using turbo.

PR #1078, opened 2026-06-04
Constrain classes further in MailPreview controller

markstory:
This controller should not attempt to load classes with `\` in the name, nor should it attempt to load classes that do not extedn `MailPreview`.

Thanks to Volker Dusch and the PHP Ecosystem security team for reporting this.

### DIFF
--- a/src/Controller/MailPreviewController.php
+++ b/src/Controller/MailPreviewController.php
@@ -23,6 +23,7 @@ use Cake\Http\Exception\NotFoundException;
 use Cake\Routing\Router;
 use Cake\Utility\Inflector;
 use DebugKit\Mailer\AbstractResult;
+use DebugKit\Mailer\MailPreview;
 use DebugKit\Mailer\PreviewResult;
 use DebugKit\Mailer\SentMailResult;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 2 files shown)


src/Controller/MailPreviewController.php
```diff
@@ -261,18 +262,21 @@ protected function findPreferredPart(AbstractResult $email, ?string $partType):
      *
      * @param string $previewName The Mailer name
      * @param string $emailName The mailer preview method
-     * @param string $plugin The plugin where the mailer preview should be found
+     * @param ?string $plugin The plugin where the mailer preview should be found
      * @return \DebugKit\Mailer\PreviewResult The result of the email preview
      * @throws \Cake\Http\Exception\NotFoundException
      */
-    protected function findPreview(string $previewName, string $emailName, string $plugin = ''): PreviewResult
+    protected function findPreview(string $previewName, string $emailName, ?string $plugin = ''): PreviewResult
```
LordSimal: Why does this need to be nullable?

ADmad: Using `null` when the mailer is not from a plugin makes sense, but then the default value of the argument should be null, not empty string.

markstory: The parameter can be null when the request doesn't have a query parameter for `plugin`. Good point about the default value.

LordSimal (approved):
This should be backported to 4.x as well and released as a bugfix

